PR #1, opened 2021-09-08
from-stream example

wesselvdv:


mikearnaldi:
Closing given it's using current, happy to re-do this with `@effect/core`

PR #10, opened 2022-10-29
Add coming from ZIO section to docs

IMax153:
cc @schickling

### DIFF
--- a/content/docs/200-coming-from-zio/index.md
+++ b/content/docs/200-coming-from-zio/index.md
@@ -1,0 +1,87 @@
+---
+title: Coming from ZIO?
+---
+
+If you are coming to Effect from ZIO, there are a few differences to be aware of.
+<br/>
+
+## Representing the Environment (`R`)
+
+In Effect, we represent the environment required to run an `Effect` workflow as a union of services:
+
+```ts twoslash
+import type { Effect } from "@effect/core/io/Effect"
+
+//                    v--------v---- `R` is a union of Console | Logger
+type Http = Effect<Console | Logger, IOError | HttpError, Response>
+
+type Response = Record<string, string>
+
+interface IOError {
+  readonly _tag: "IOError"
+}
+
+interface HttpError {
+  readonly _tag: "HttpError"
+}
+
+interface Console {
+  readonly log: (msg: string) => void
+}
+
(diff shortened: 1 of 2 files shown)


mikearnaldi:
As we noticed already errors are not showing properly:

Screenshot 2022-11-01 111830

Also we speak about R being phantom, a phantom parameter is a parameter that is not used but in the examples to try to explain the "why" you use examples that try to use the parameter.

Namely:

```ts
// @errors: 2322
interface A {
  readonly prop: string
}
interface B {
  readonly prop: number
}
const ab: A & B = {
  prop: ""
}
```

and 

```ts
interface A {
  readonly prop: string
}
interface B {
  readonly prop: number
}
const ab: A | B = {
  prop: ""
}
```

A batter more simple example would be:

```ts
interface ConsoleService {
  readonly _tag: "Console"
}
interface LoggerService {
  readonly _tag: "Logger"
}
type Env1 = ConsoleService & LoggerService 
//     ^?
//     never because of the conflict between "Console" and "Logger" as there is no string which is both "Console" and "Logger"
type Env2 = ConsoleService | LoggerService 
//     ^?
//     "ConsoleService | LoggerService" as there is no problem with a value that can be either Console or Logger
```

Further examples in using unions to represent intersection of non conflicting values can be the fact that:

```ts
interface ABC {
  a: 0
  b: 1
  c: 2
}
type Key = keyof ABC
//     ^?
//     a | b | c
```

IMax153:
The purpose of the example was to demonstrate as simply as possible what happens when an intersection tries to unify un-unifiable types. 

If we want to make the example more complex to keep with the environment theme that's fine with me. 

And the error rendering has something to do with the last few lines in the custom CSS file for Shiki Twoslash. Specifically, if you comment out the lines related to GitHub styles, the issue goes away.

mikearnaldi:
Given we are talking about environment I would opt in for a more verbose but less confusing example, even because the example shown doesn't show the problem if TS would only reduce properties to "never" it would be totally fine to use as we are basically using the parameter nominally the problem is that in some cases conflicts (namely conflicts on literal types) end up reducing to never the parent (in the example case `ConsoleService & LoggerService` itself and not only a property of those)

PR #11, opened 2022-12-19
Docs Week

mikearnaldi:


content/docs/200-what-is-effect/the-effect-type.md
```diff
@@ -0,0 +1,26 @@
...
+which have the following meaning:
+
+- `Context`: Represents the contextual data which is required by the `Effect` in order to be executed
+
+- `Failure`: Represents _expected_ failures that can occur when an `Effect` is executed
+
+- `Success`: Represents the type of the value that an `Effect` can succeed with when executed
+
+<Callout>
+Throughout the Effect ecosystem, you will often see `Effect`'s type parameters abbreviated as `R`, `E`, and `A` respectively. This is just shorthand for `R`equirements, `E`rror, and the success value of type `A`.
```
cjol, line 20: This line uses terminology inconsistently with the line above. Is it `R`equirements or `Context`? `Failure` or `Error`? `Success` or ...`A`Output? 

On that note, can we think of some other mnemonic that begins with the letter `A`, or should we be explicit that the letter `A` is used as a historical convention rather than having any meaning. Indeed, are we happy with "historical convention" as a basis for a design choice, or should we consider using something like `Effect<R,E,O>` at this stage?

content/docs/200-what-is-effect/the-effect-type.md
```diff
@@ -0,0 +1,26 @@
+---
+title: The Effect Type
+---
+
+Effect was designed from the ground up to be a completely type-safe representation of any program.
```
cjol, line 5: It would be great to have a short paragraph here about what the benefit is of having a "representation of a program" (i.e. that it can be composed and manipulated in powerful ways to build up complexity through composition instead of by adding complexity to our application logic)

PR #14, opened 2023-02-11
[WIP] docs: early draft of something about error handling

aniravi24:
This was originally a "Why Effect" page, but it ended up being too specific to error handling so going to mark it as such. Might scrap this page, or find some use for it elsewhere in the docs.

### DIFF
--- a/pages/docs/core-concepts/batching-caching.mdx
+++ b/pages/docs/core-concepts/batching-caching.mdx
@@ -338,7 +338,10 @@ interface GetUserByIdCache {
   readonly _: unique symbol
 }
 
-const GetUserByIdCache = Context.Tag<GetUserByIdCache, Request.Cache<GetUserById>>()
+const GetUserByIdCache = Context.Tag<GetUserByIdCache, Request.Cache<GetUserById>>(
+  // this is not strictly needed but makes sure the tag is the same if module is reloaded
+  Symbol.for("@app/services/GetUserByIdCache")
+)
 
 const GetUserByIdCacheLive = Layer.effect(
   GetUserByIdCache,


aniravi24:
Maybe this page tries to be too concrete. Prisma's "why Prisma" doc is all high-level but makes clear points

https://www.prisma.io/docs/concepts/overview/why-prisma

aniravi24:
Looking at this again, I think that this is probably too specific. The Effect ecosystem does a lot of different things and I don't want someone to jump into this thinking the main point is better error handling, because it's more than that. Something like this could probably live elsewhere in the docs?

Maybe this page could be a slightly expanded version of @schickling's tweets on Effect: https://twitter.com/schickling/status/1619052200889835520?cxt=HHwWgIDTrcHZg_gsAAAA

cjol:
> Looking at this again, I think that this is probably too specific. The Effect ecosystem does a lot of different things and I don't want someone to jump into this thinking the main point is better error handling, because it's more than that. Something like this could probably live elsewhere in the docs?
> 
> Maybe this page could instead be a slightly expanded version of @schickling's tweets on Effect: https://twitter.com/schickling/status/1619052200889835520?cxt=HHwWgIDTrcHZg_gsAAAA

I think you might be right, and the error-handling content could be in a dedicated page. I think the motivation page should focus on what an "effect is" (a description of action rather than the action itself) and the benefits that having such a description brings through being able to manipulate that description (error handling, retrying, aborting, tracing, etc). 

Each of those types of manipulation could have a brief concrete example of the problem that Effect solves, then link through to a dedicated page which shows how that part of the Effect toolbox addresses the problem.

Finally you could wrap up by demonstrating that because al these solutions are backed by sound mathematical theory, they compose in an elegant way, such that tackling every element of complexity listed above doesn't exponentially increase the complexity of your overall code.

aniravi24:
I'm going to change the description to indicate that this page will probably have something to do with error handling instead, but not sure if we should keep it.

content/docs/101-why-effect/index.md
```diff
@@ -0,0 +1,151 @@
...
+<br />
+For the following set of examples, observe how the structure of the code changes, and don't worry about how we implement a specific action like sending an email.
+<br />
+
+## Simple program (no error handling)
+
+Let's imagine a simple TypeScript program that creates a user in the database and invites a user via email.
+
+```ts twoslash
+declare function createUserInDB(args: { name: string; email: string }): { id: string; name: string; email: string }
```
ahrjarrett, line 19: Should `createUserInDB` and `sendUserInvite` return Promises?

content/docs/101-why-effect/index.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+Now, we've done a basic form of "error handling" by wrapping everything in a `try/catch`. This is the easy way to do it, and it's surprising how often code is left like this. The only difference between this and the previous example is that we prevented our app from crashing, but we didn't actually do anything to handle the errors that each function can throw. When our entire codebase looks like this, it becomes really hard to make sense of what's happening in our codebase when things go wrong.
+<br />
+
+There's also some additional problems here:
+
+// TODO these docs don't handle numerical lists correctly
+
+    1. We have no idea what kinds of errors could be thrown by `createUserInDB` or `sendUserInvite`. I just have to assume that they will throw exceptions and always wrap my code in a `try/catch`.
```
ahrjarrett, line 60: Disregard this if you do this later on, but it might help drive the point home if you explicitly point out that TypeScript does natively provide a way to express that a function might throw -- which effectively means you cannot reason about how your program will behave at runtime from the signatures alone.

That makes any program's behavior impossible to predict without "breaking open" each and every function and reading the entire implementation.

Which seems to be what you're getting at here -- in the past when I've discussed this subject this teammates, the fact that TypeScript does not have an answer for this (and that nobody talks about it) does seem to pique people's interest

content/docs/101-why-effect/index.md
aniravi24: You're right, given how I'm using it, technically should be a Promise

content/docs/101-why-effect/index.md
aniravi24: I agree, I've had the same epiphany when talking to teammates about this subject. If I end up tweaking this or including this in the docs in some way, will definitely make sure to mention this.

aniravi24:
Was syncing with main and decided to close this for now, can put the content back in a different form later if needed

PR #16, opened 2023-02-12
WIP: First draft of a "Taming Complexity with Effect" guide

cjol:
I've been offline and writing stream-of-consciousness. I haven't re-read
through to check for narrative structure, consistency or even correctness, so
I've marked as draft until I can come back to it. I know for a fact that I
haven't provided enough information for twoslash to work it's magic, for example!

I have stubbed out plans for future expansion and have no ego if anyone wants
to step in and continue before I get a chance to pick up again! It's almost
certainly worth splitting this into multiple pages before going much further.

I've changed the sequencing slightly compared to the WorkerConf talk, because I
think we can do a better job of introducing concepts one-by-one, but I'm not wedded to this sequence
by any means.

Also, a question (which I might move to Discord once I'm back online): what's
the "official" name for this library? In my writing I'd like to use
capitalisation consistently to help with understanding, i.e.:

- `effect` is a common noun, as in "This function has the effect of reading from an API"
- `Effect` is the name of the type, as in "This function returns an `Effect`."

And I'm not sure how to capitalise the name of the library in order to
distinguish from both of these uses. I'd love to call it Effect-TS or something
for clarity, but that's not what it's called! An example sentence might be "Our
function reads from an API, so we represent that effect with Effect's `Effect`"

### DIFF
--- a/content/docs/700-taming-complexity/index.md
+++ b/content/docs/700-taming-complexity/index.md
@@ -1,0 +1,416 @@
+---
+title: Taming Complexity with Effects
+---
+
+Writing software is easy. Unfortunately, "the real world" constantly likes to
+get in our way, introducing additional complexity that makes it hard for us to
+do what we want to do.
+
+<!-- If we can do Twitter embed, see @kentbeck's famous tweet on refactoring to make the hard thing easy -->
+
+Let's say we have a revolutionary new app idea: a todo list app. Our core
+application logic is simple -- it just fetches TODO items from a JSON API.
+
+```ts twoslash
+/// <reference path="node_modules/@types/node/index.d.ts" />
+// ---cut---
+interface TodoItem {}
+
+const getTodoPromise = async (id: number): Promise<unknown> => {
+  const res = await fetch(`https://jsonplaceholder.typicode.com/todos/${id}`)
+
+  return (await res.json()) as TodoItem
+}
+```
+
+See -- easy! But... this won't take us very far. There is a whole raft of "real world" complexity that we're not handling:
+
+- How can we efficiently get multiple todos at once?
+- What if the network goes down or the API misbehaves?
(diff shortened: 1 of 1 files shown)


jessekelly881:
This was a great read!

content/docs/700-taming-complexity/index.md
```diff
@@ -0,0 +1,416 @@
...
+in an error of type `Error` instead. We're also documenting that `getTodo` has
+no "Requirements", so our consumer can invoke the effect without providing
+anything more (we'll revisit that when we look at dependency injection later).
+
+Keep reading to find out how this simple abstraction can be powerful enough to
+tackle all the complexity described above!
+
+# Concurrency
+
+In the real world, we need to exploit concurrency. Fetching todos one-at-a-time
```
mikearnaldi, line 72: may be worth to avoid "exploit" in favour of "leverage", I am personally not bothered but I know some interpret the terms in a non positive way

content/docs/700-taming-complexity/index.md
```diff
@@ -0,0 +1,416 @@
...
+even needs error-handling in the first place.
+
+To access this propagated error, we can use `Effect.catchAll`. This function is
+a kind of an inverse to `flatMap`: it will only call the provided callback if
+the previous `Effect` was _not_ successful.
+
+```ts
+// the implementation is not important for now -- but note that the Result type is the same
+// as `getAndSaveTodo` (string), while the Failure type is different (never vs `Error`)
+const gracefullyRenderError = (x: Error) => Effect.succeed('Error Message Displayed')
```
mikearnaldi, line 257: may be worth using `Effect.logError` in those scenarios (or mention it)

mikearnaldi:
Looks great so far!

PR #29, opened 2023-04-12
Hover issue solved

coleea:
As shown in the screenshot below, the issue of not being able to see the contents of hover has been resolved

Screenshot_2023-04-11-19-28-48-59_40deb401b9ffe8e1df2f1cc5ba480b12

This now looks like the picture below

effect-official-website vercel app_docs_core-concepts_error-handling

This is implemented in `root_folder/hooks/useCustomStyle.ts`

I'll fix it if there's a problem


IMax153 (approved):

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,148 @@
...
+        elementStaticXPos,
+        elementScrrenXPos,
+        contentWidth,
+        allowedContentFullWidth,
+        elementScrrenXPosRatio,
+        padding
+    }
+}
+
+export function useCustomStyle() {
```
datner, line 80: This seems like a ton of work just to always position a tooltip a static amount of pixels from the left of it's relative container in mobile.. The same-or-similar behavior can be achieved by a simpler mechanism. Something similar to the below maybe?
```ts
const styles = getComputedStyle(hint, ':before')
const width = parseInt(styles.width.replace('px',''))
const left = parseInt(styles.left.replace('px',''))
const maxWidth = codeContainer.getBoundingRect().width
const moveLeftBy = Math.min(0, maxWidth - width - left)
```
You can reach out in discord if you want to talk more about it

datner (changes requested):
I think we can do this even better, but I'm also starting to doubt the value of this change in the grand scheme of things 🤔

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,148 @@
...
+function isMobile() {
+  const userAgent = navigator.userAgent || navigator.vendor ;
+  return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(userAgent);
+}
+
+// const isMobile = () => /iPhone|Android/i.test(globalThis.navigator.userAgent);
+
+function getPadding(mainElement: HTMLElement) {
+  return Number(getComputedStyle(mainElement).paddingLeft.replace(/[^0-9]/g, "")) ;
+}
```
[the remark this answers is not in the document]
coleea: when I use computedStyleMap, the error occurs
Property 'computedStyleMap' does not exist on type 'HTMLElement'.ts(2339)
I managed to do using as any
(currentElement as any).computedStyleMap().get('padding-left').value

components/TwoslashPatch.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    left: ${allowedContentFullWidth - contentWidth - (padding * 2) }px !important;
+                   }`
+                :`pre.twoslash data-lsp:hover::before {
+                    left: ${elementScrrenXPos}px !important;
+                  }`)
+                }
+            </style>
+        , document.body)}
+    </>
+}
```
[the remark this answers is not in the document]
coleea: I don't know what exactly "the manual manipulation of the dom" means. 
I fix the code like this

```typescript
export default function TwoslashPatchPortal() {

    const {
        isMounted,
        currentStyle
    } = useCustomStyle()

    return  <>
        {isMounted && 
        createPortal(
            <style>                
                {currentStyle}
            </style>
        , document.body)}
    </>
}


```

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,148 @@
...
+function getPseudoElementWidth(currentElement: HTMLElement, pseudoName : string) {
+
+    const computedStyle = getComputedStyle(currentElement, pseudoName);
+
+    // dynamically create element because of computing content width dynamically
+    const customCssElement = document.createElement('div');
+    customCssElement.style.position = 'absolute';
+    customCssElement.style.visibility = 'hidden';
+    customCssElement.style.pointerEvents = 'none';
+    customCssElement.className = 'invisible-div';
```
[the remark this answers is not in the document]
coleea: className is not needed. will delete this.
.visibility = 'hidden'; is  for not showing this element because this element is only for computing width of element
.pointerEvents = 'none'; is for preventing the shape of the cursor changed

hooks/useCustomStyle.tsx
coleea: you are right. this is better

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,148 @@
...
+      'text-align',
+      'padding',
+      'border-radius',
+      'font-family',
+      'white-space',
+      'margin',
+    ];
+
+    stylesToCopy.forEach(style => customCssElement.style[style as any] = computedStyle.getPropertyValue(style));
+
```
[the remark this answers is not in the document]
coleea: I remove this code accoring to apply strategy you mentioned below

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,148 @@
+import { useEffect, useState } from "react";
+import { createPortal } from "react-dom";
+
+function isMobile() {
+  const userAgent = navigator.userAgent || navigator.vendor ;
+  return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(userAgent);
+}
```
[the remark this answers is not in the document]
coleea: matchMedia is applied

patch.patch
```diff
@@ -0,0 +1,219 @@
+diff --git a/hooks/useCustomStyle.ts b/hooks/useCustomStyle.ts
```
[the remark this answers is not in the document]
coleea: removed

components/TwoslashPatch.tsx
```diff
@@ -0,0 +1,46 @@
...
+        contentWidth,
+        allowedContentFullWidth,
+        elementScrrenXPosRatio,
+        padding
+    } = useCustomStyle()
+
+    console.log({contentWidth});
+    console.log({allowedContentFullWidth});
+    console.log('contentWidth/allowedContentFullWidth');
+    console.log(contentWidth!/allowedContentFullWidth!);
```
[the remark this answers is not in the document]
coleea: removed

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+function isInCodeContainer(currentElement : HTMLElement) {
+    const containerElement = currentElement.closest('.code-container')!
+    return containerElement?.scrollLeft
+}
+
+function getValuesForCSSLeft( currentElement : HTMLElement) {
+
+    const screenWidth = globalThis.window.outerWidth;
```
datner, line 24: `window` is normally accessed directly. 
Also, iirc, `window` does not have any border width or padding, which is why `window.innerWidth` is more commonly used.

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const allowedContentFullWidth = screenWidth - (mainLeftPadding * 2)
+
+    const elementXPos = currentElement.getBoundingClientRect().x - mainLeftPadding
+    const elementXPosRatio = elementXPos / allowedContentFullWidth
+
+    return {
+        elementXPos,
+        allowedContentFullWidth,
+        elementXPosRatio,
+    }
```
datner, line 37: can you please explain to me the logic in this function? I don't quite understand why the window and  the`<main/>` are related to the calculations required... 🤔

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,122 @@
...
+      if(isMounted ) {
+          window.addEventListener("touch", touchAndClickListener)
+          window.addEventListener("click", touchAndClickListener)
+      }
+
+      return () => {
+          window.removeEventListener("touch", touchAndClickListener)
+          window.removeEventListener("click", touchAndClickListener)
+      }
+  }, [isMounted]);
```
datner, line 116: `useEffect` will always run a _after_ a component has mounted, meaning that `isMounted` actually checks for the **2nd** render of a component and not the mount of the component.

This effect violates the rules of hooks, `touchAndClickListener` is _not_ a static or a stable function, therefore it should be in the deps array.
But more likely the flow control here need to change in general to move the function inside the `useEffect`

hooks/useCustomStyle.tsx
```diff
@@ -0,0 +1,122 @@
...
+            allowedContentFullWidth,
+            elementXPos,
+        })
+    )
+  }
+
+  useEffect(() => {
+      setIsMounted(true);
+      return () => setIsMounted(false);
+  },[])
```
datner, line 99: the flow here is as follows:
```
mount 
-> effect (setsMount) [trigger rerender] 
-> effect (add listeners -- not mounted )
-> rerender
-> effect {deps changed} (add listeners -- is mounted)
```
not really sure what is the point..

`useEffect` _runs only in the client side_ and not on SSR or SSG, meaning that it is safe to access the `window` object inside an effect. This mount check is redundant

components/TwoslashPatch.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function TwoslashPatchPortal() {
+
+    const {
+        isMounted,
+        currentStyle
+    } = useCustomStyle()
+
+    return  <>
+        {isMounted &&
```
datner, line 13: the initial render is the mount, this conditional is redundant. It can only reach this line to check `isMounted` _if it's mounted_. This boolean is actually saying `isTheSecondRenderOrLater`

datner (changes requested):
We're almost there, lets go over it one last time and get this in!
Lets try to inline things before we extract them. 
In my opinion the following:
```ts
function isMobile() {
  return matchMedia(...).matches
  }

if (isMobile()) {...}
```
is not _more_ readable, but is more noisy than
```ts
const isMobile = matchMedia(...).matches
if (isMobile) {...}
```
There are a few instances of these, the logic we're expressing here can be even simplified further.

Also, please use the included formatter in the project (either `prettier` or `dpring` I don't recall) 🙏🏻

hooks/useCustomStyle.tsx
coleea: opinion about isMobile <- applied
prettier  <- applied
window.innerWidth <- applied

hooks/useCustomStyle.tsx
coleea: removed "isMounted" from Dependency Array

removed "if(isMounted)"

"touchAndClickListener function" moved inside the useEffect

hooks/useCustomStyle.tsx
coleea: change the name "isMounted" to "isClientReady"

the logic in useEffect no longer depends on this

components/TwoslashPatch.tsx
coleea: change the name "isMounted" to "isClientReady"

I don't want to use "useState" for checking is SSR or CSR. I did something various but It is the only way to prevent the error `Error: Hydration failed because the initial UI does not match what was rendered on the server.`

hooks/useCustomStyle.tsx
coleea: KakaoTalk_20230421_190809795

PR #32, opened 2023-04-12
feat: migrated docs from @effect/data

jessekelly881:
Migrated docs from @effect/data and added Mermaid rendering. Unfortunately I had to use a react component for mermaid rendering. I tried several using remark plugins for mermaid with no success. :(


IMax153 (changes requested):
Thanks @jessekelly881 ! A few minor comments:

1. Can we make the capitalization here consistent? Also `Algebraic operations` should probably not be nested.

image

2. I think we are missing some peer deps. Specifically `mermaid` is causing an error for me.

image

jessekelly881:
Fixed the dep. issue. Actually, I also got mermaid rendering to work without using an inline component as well. What's the issue with capitalization though? It looks correct to me.

IMax153:
> What's the issue with capitalization though? It looks correct to me.

It's just a nitpick really - I'd like everything in headings to be in [title case](https://en.wikipedia.org/wiki/Title_case) but I can always go back and do a pass to fix all this later on.

Will pull down your changes and test locally soon 👍 Thanks for addressing!

jessekelly881:
Oh! Now I understand! I thought you meant consistency in what word was capitalized. jaja. I'll update them.

IMax153:
Just pulled the changes and tested. They look good to me. Once you've finished modifying the titles ping me again and I'll approve / run CI / merge.

steida:
https://title.sh/ FTW

jessekelly881:
Thanks @steida that site came in useful. @IMax153 updated.

IMax153:
@jessekelly881 - looks like there are some `pre-commit` hooks failing. I'd suggest installing the hooks in your project locally with:

```bash
pre-commit install
```

Note that this assumes you have `pre-commit` installed on your machine.

Once installed, you can run the hooks manually with:

```bash
pre-commit run --all-files
```

jessekelly881:
Yeah? I'll take a look. Btw, in general, having pre-commit hooks that ship with a project is generally a bad idea. It's much better to run tests as part of ci/cd(e.g. as part of GitHub workflows)

IMax153:
> Btw, in general, having pre-commit hooks that ship with a project is a bad idea. It's much better to run tests as part of ci/cd(e.g. as part of GitHub workflows)

@jessekelly881 - I disagree with this sentiment entirely. The entire point of `pre-commit` hooks is to enforce code style / formatting / etc. Having them run on commit when a developer is working is the entire point.

We also have a CI-CD step in our GitHub actions that runs our `pre-commit` steps to check that everything is conformant to the rules we've specified and to catch situations such as this.

Furthermore, if a developer working on our projects has a full `nix` / `direnv` setup then the `pre-commit` hooks are setup automatically as part of the `nix` dev shell.

jessekelly881:
Yeah, I understand the rational. I've used them for exactly this reason in the past. The problem is that it's a bit too invasive to a developers local "flow". It effects my local version of the project. If I want to create badly formatted unworking commits in my local(or commit on every key press) and then run git rebase -i once I actually want to submit that change. Those checks shouldn't be prescribed by the project. The only thing that's relevant from the remotes point of view is the final pr.

IMax153:
You can just `--skip-verify` if you want to commit without running hooks.

jessekelly881:
You can.. but that's not quite the point. Also, in some editors like vscode the UI commit will fail and as far as I have discovered there isn't any way around it. You have to drop into a terminal. Also, they just aren't very useful in my opinion.. especially for projects with multiple contributors. You can't rely on the user to run them so you have to run the tests on the remote anyway.

IMax153:
I think we can agree to disagree :). @mikearnaldi and I have made the decision to use `pre-commit` to enforce consistent style and formatting _precisely because_ these are projects with multiple contributors.

jessekelly881:
Jajaja. That's fine. :) It's a very minor thing anyway. It's opt in on your local machine anyway. Any reason you guys went with pre-commit instead of husky btw?

IMax153:
> Any reason you guys went with pre-commit instead of husky btw?

We've used `husky` in the past but didn't like it much.

jessekelly881:
Yeah? Also, totally bias video supporting my point jajaja. https://youtu.be/RAelLqnnOp0

IMax153:
We didn't have issues, we just prefer using pre-commit. I'm sure I could find an equally compelling video touting the benefits of `pre-commit`. Also, I generally don't take advice from influencers as I prefer to formulate my own opinions about tooling through experience.

The point is given that Mike and I are maintainers of the project, we've decided to use this tool to help enforce formatting and styling to help make our lives a bit easier. We like this tool so we chose to use it. Does that mean we are married to this tool? No. May we choose to remove it in the future? Yes, just like we removed `husky`. But for now it stays.

I apologize if it is an inconvenience to your local workflow.

IMax153:
Hey @jessekelly881 - just wanted to let you know I made a few minor commits to get this PR over the finish line. We appreciate your contribution very much! Thank you very much for your work getting these docs ported into the website!

IMax153 (approved):

jessekelly881:
Thanks @IMax153. I normally would have fixed it sooner but I've been pretty under the weather for the last week. :( I'm glad we could get these docs into the website.

PR #41, opened 2023-04-16
feat: Why Effect page

aniravi24:
Starting to provide a mental model for how someone can think about using Effect and what it enables for them.

More iterations could be done on this to make it narrative driven with examples in the future!

### DIFF
--- a/pages/docs/why-effect.mdx
+++ b/pages/docs/why-effect.mdx
@@ -26,7 +26,7 @@ Often, we will do the "easiest" thing and just wrap the function in a `try/catch
 
 While Effect is a vast ecosystem of many different tools, if it had to be reduced down to just one unique idea, it would be the following:
 
-Effect's major unique insight is that we can use the type system to track _errors_ and _"dependencies"_ (more on this later), not only _success_ values as shown in the divide example above.
+Effect's major unique insight is that we can use the type system to track _errors_ and _"context"_ (more on this later), not only _success_ values as shown in the divide example above.
 
 Here's the same divide function from above, but with the Effect pattern:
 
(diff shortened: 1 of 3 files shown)


pages/docs/why-effect.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+Errors now become values, just like your success values. Effect gives us many functions to make managing errors and success values ergonomic.
+
+Additionally, tracking dependencies in your types allows you to write modules separate from your implementation detail. For example, your core domain logic (e.g. creating a user account) can depend on some external API (e.g. sending an account invite email), and you can write that API as an interface/service (e.g. email service). When you write tests, you can easily swap out the API for a mock implementation (e.g. mock email service) and test your core logic without having to worry about the implementation details of the API, and without changing any of your core logic. Don't worry if this doesn't make sense yet. It's one of Effect's most powerful concepts that you will learn more about as you go through the [Core Concepts](../core-concepts/effect).
```
aniravi24, line 48: This is my least favorite section - I didn't immediately have an intuition for how to describe this better. 

This is one of my favorite Effect features but I only started understanding it after using it so I have a hard time describing it generally without an example.

mikearnaldi: I wouldn't approach it from the perspective of tracking dependencies, this specific framing in a general sense is called "effect tracking" and, while loved by many, it is considered "useless" for business by many others (including myself). What is kind of generally agreed as being valuable is having a computation which is context dependent, for example the reasons you'd use a context in react is to provide external information to a component in the same way the effect context is used to provide information to a program - the provided information can be dependencies / flags / configs / caches / and really anything

mikearnaldi:
It also may be worth including parts of the efforts in https://github.com/Effect-TS/website/pull/16 I thought that overall was a good read

pages/docs/why-effect.mdx
aniravi24: Thanks for this feedback! I have a better sense of how to frame it now. Will revisit

aniravi24:
> It also may be worth including parts of the efforts in #16 I thought that overall was a good read

I also like it as well, trying to think through how and where to include it (or if should potentially be a separate page).

aniravi24:
> It also may be worth including parts of the efforts in #16 I thought that overall was a good read

This could be a separate page or a guide, I like it standalone. Either that should replace my page here if we prefer the narrative explainer with an example or we can have both - a shorter explainer like mine and a larger guide in the same section walking through a narrative-driven example of how to think through the challenge of working on something.

pages/docs/why-effect.mdx
```diff
@@ -0,0 +1,71 @@
+# Why Effect?
+
+## Motivation
+
+Programming is challenging. When we build libraries and apps, we look to many tools to handle the complexity and make our day-to-day more manageable.
+
+One of the most important tools we have in TypeScript is the compiler. It is the first line of defense against bugs, domain errors, and general complexity. Effect presents a new way of thinking about programming in TypeScript.
```
IMax153, line 7: I would suggest moving the statements about the TypeScript compiler down below in the paragraph that starts with `In "typical" TypeScript`. And then the final sentence of this paragraph can become the last sentence of the first paragraph.

IMax153 (approved):
Overall - I think this page is written extremely well! Kudos to @aniravi24 and thank you for working on it! I don't have any other major comments other than the one in my review and what Mike already mentioned.

pages/docs/why-effect.mdx
aniravi24: Done, thanks!

aniravi24:
> It also may be worth including parts of the efforts in #16 I thought that overall was a good read

Let's merge this for now, and happy to discuss in the other PR how we could integrate that into the docs (or potentially replace this page if we want). I think it lives nicely on its own outside of this page!

PR #42, opened 2023-04-16
feat: put something on the Getting Started page

aniravi24:


pages/docs/getting-started.mdx
```diff
@@ -1,3 +1,23 @@
...
+  Effect's docs are still very much a work in progress!
+</Callout>
+
+## Start Reading
+
+If you'd like to read more about Effect, start reading through the [Core Concepts](../core-concepts/effect.mdx)
+
+## Crashcourse
+
+If you're looking to dive right in, check out the [unofficial crashcourse](https://github.com/pigoz/effect-crashcourse/)
```
aniravi24, line 15: Not sure if we wanted to link to this (or if we might even consider making this an "official" repo, just put this here for now as its the only way I would currently recommend anyone getting started with Effect. Let me know if this is fine

datner: !!not an authority!!
I think this is fine, it's not as if theres an alternative and this crashcourse has proved invaluable

mikearnaldi: I am not against linking it but what would be great is to absorb it, I know @pigoz planned to migrate it to the docs already, maybe a good opportunity

IMax153: Yea honestly I think the less places an individual learning Effect has to "travel" to in order to get information, the better. 

I agree with @mikearnaldi that this is probably a good time to get content from the crash course into the docs site.

IMax153:
Another thought, under the `Getting Started` section perhaps we can have a sub-page which discusses the `@effect` vs `effect` modules on NPM and how to install/use them.

pages/docs/getting-started.mdx
aniravi24: I agree that preferably we have it directly on the site. Where do you see the content going? Should it be directly incorporated into core concepts, or would it be a separate part of the docs?

aniravi24:
> Another thought, under the `Getting Started` section perhaps we can have a sub-page which discusses the `@effect` vs `effect` modules on NPM and how to install/use them.

I'll definitely incorporate this somehow. People are confused by this often (understandably so) while new Effect is in this transition period.

pages/docs/getting-started.mdx
aniravi24: Could also put it under a section called crashcourse or overview or something, and then re-organize later.

aniravi24:
> Another thought, under the `Getting Started` section perhaps we can have a sub-page which discusses the `@effect` vs `effect` modules on NPM and how to install/use them.

I added a small stub at the top of this of this file for now, can polish or move elsewhere later.

pages/docs/getting-started.mdx
IMax153: @aniravi24 - I think it should be directly included under core concepts. A lot of the content I already wrote there was inspired / ported from the crash course 😅 .

IMax153:
> I added some copy mainly as a placeholder for this at the top of this file, for now, can polish or move elsewhere later.

I think this is fine for now, we can always add detail over time. The point is to just get some reasonable content out there (so thank you for helping with that :) )

pages/docs/getting-started.mdx
aniravi24: Ok sure, let me remove the crashcourse link for now and we can get this merged? Can do a separate PR on GitHub for starting working on a port if we decide to proceed there @IMax153

IMax153 (approved):

PR #43, opened 2023-04-17
fix: getting started links pointing to wrong location

aniravi24:


aniravi24:
@IMax153 super simple fix so I'm just going to merge, hope that's fine!

PR #48, opened 2023-04-22
docs(core-concepts): page on services

aniravi24:


aniravi24:
Some things that aren't included yet:
1. Passing in a string or symbol to `Context.Tag`
2. The two arguments to the Tag type constructor (separating the identifier from the actual service)

pages/docs/core-concepts/services.mdx
```diff
@@ -0,0 +1,165 @@
...
+Let's create a service that implements a random number generator.
+
+```twoslash include service-interface
+import * as Effect from "@effect/io/Effect"
+import * as Context from "@effect/data/Context";
+import { pipe } from "@effect/data/Function";
+
+
+export interface CustomRandom {
+  readonly next: () => number;
```
mikearnaldi, line 20: this should be:

```ts
interface CustomRandom {
  readonly next: Effect.Effect<never, never, number>
}
```

as random number generation is an effect, in a pure world `() => number` can only return a constant

aniravi24: Sure, may opt for a different example entirely in the future but we'll go with this for now.

PR #49, opened 2023-04-22
docs(core-concepts): add information on how to actually run an Effect

aniravi24:


pages/docs/core-concepts/_meta.json
```diff
@@ -1,5 +1,5 @@
 {
-  "effect": "Effect",
+  "create-and-run-effects": "Create and Run Effects",
```
IMax153, line 2: Maybe let's just name this to `creating-effects` and we can include examples of running the Effects themselves within the documentation.

pages/docs/core-concepts/create-and-run-effects.mdx
```diff
@@ -44,9 +46,9 @@ In the sections below, we walk through some of the more common use cases.
 To create an `Effect` from a value, you can use the `succeed` constructor:
 
 ```ts twoslash
-import * as Effect from "@effect/io/Effect"
+import * as Effect from "@effect/io/Effect";
```
IMax153, line 49: I would prefer we leave the semicolons out of all code samples - let's go ahead and remove them everywhere. The less syntax noise the better.

pages/docs/core-concepts/create-and-run-effects.mdx
```diff
@@ -106,48 +107,99 @@ Once again inspecting the return type we can see that it is `Effect<never, never
...
-  NodeFS.readFile("todos.txt", (error, data) => {
-    if (error) {
-      resume(Effect.fail(error))
-    } else {
-      resume(Effect.succeed(data))
-    }
-  })
-})
+export const readTodos = Effect.async<never, NodeJS.ErrnoException, Buffer>(
+  //              ^?
```
IMax153, line 123: Similar issue here with spacing the twoslash comment.

pages/docs/core-concepts/create-and-run-effects.mdx
```diff
@@ -106,48 +107,99 @@ Once again inspecting the return type we can see that it is `Effect<never, never
...
+      } else {
+        resume(Effect.succeed(data));
+      }
+    });
+  }
+);
 ```
 
-Something to note in the above example is that we were actually forced to manually annotate the types when calling `Effect.async`. Unfortunately, TypeScript is not capable of inferring the type parameters for a callback based on what is returned inside the body of the callback. This limitation of TypeScript is the one minor drawback of using callback-based APIs.
+Something to note in the above example is that we were actually forced to manually annotate the types when calling `Effect.async`. Unfortunately, TypeScript is not capable of inferring the type parameters for a callback based on what is returned inside the body of the callback. This limitation of TypeScript is the one minor drawback of using callback-based APIs. However, it will validate that the types you passed in match what is being returned.
```
IMax153, line 136: I would maybe change this to say: "Annotating the types of `Effect.async` will ensure that the values provided to `resume` match the expected types.".

pages/docs/core-concepts/create-and-run-effects.mdx
```diff
@@ -106,48 +107,99 @@ Once again inspecting the return type we can see that it is `Effect<never, never
...
-  })
+    });
+  });
+
+export const readTodos = readFile("todos.txt");
+//           ^?
+```
 
-export const readTodos = readFile("todos.txt")
+## Running Effects
```
IMax153, line 163: In all honesty, we could consider moving this to a dedicated section since there are various ways of running Effects and they all have slightly different implications. 

I think it may be better to go ahead and do that - let's create a section called "Running Effects" in the sidebar.

IMax153 (changes requested):

pages/docs/core-concepts/create-and-run-effects.mdx
steida: @IMax153 The only case when semicolons are visible is when someone explicitly mentions them. Or when the lack of them breaks prettier.

IMax153: Is this not considered "explicitly mentioning them"?

pages/docs/core-concepts/create-and-run-effects.mdx
aniravi24: I was considering this as well, I initially didn't have enough content so I put them in the same place, but I think we can separate them and add content later.

pages/docs/core-concepts/create-and-run-effects.mdx
aniravi24: Thanks!

pages/docs/core-concepts/create-and-run-effects.mdx
aniravi24: Anything else you think we should add in here at this second? Or should we merge and revisit?

pages/docs/core-concepts/running-effects.mdx
```diff
@@ -0,0 +1,43 @@
...
+```ts twoslash
+import * as Effect from "@effect/io/Effect"
+
+// sync variant
+export const programSync = Effect.runSync(Effect.succeed("Hello, World!"))
+//           ^?
+
+// promise variant
+export const programPromise = Effect.runPromise(
+  //           ^?
```
IMax153, line 16:
```suggestion
export const programPromise = Effect.runPromise(
//           ^?
```

pages/docs/core-concepts/running-effects.mdx
```diff
@@ -0,0 +1,43 @@
...
+  Effect.succeed("Hello, World!")
+)
+
+// promise Either variant
+export const programPromiseEither = Effect.runPromiseEither(
+  //           ^?
```
IMax153, line 22:
```suggestion
export const programPromiseEither = Effect.runPromiseEither(
//           ^?
```

pages/docs/core-concepts/running-effects.mdx
```diff
@@ -0,0 +1,43 @@
...
+  Effect.succeed("Hello, World!")
+)
+
+// callback variant
+export const programCallback = Effect.runCallback(
+  //           ^?
```
IMax153, line 28:
```suggestion
export const programCallback = Effect.runCallback(
//           ^?
```

IMax153 (changes requested):
Just a few minor requested changes, otherwise the content looks great 👍

pages/docs/core-concepts/running-effects.mdx
aniravi24: I believe prettier will put these back to where they were automatically. Might have to leave these as is.

aniravi24: Yeah this is one of the caveats I mentioned in my other PR. Prettier will format all comments to align with your code (mainly the code underneath the comment). There's no way I'm aware of to turn that off, so you just have to make sure the `?` is aligned correctly after prettier formats and you're good.

IMax153: This isn't an ideal solution, but we could try:
```
{/* prettier-ignore */}
```ts
```

aniravi24: I wouldn't want to ignore formatting the entire code block just because of the comment - I don't think the comment formatting is that big of a deal, considering that everything else "just works" with very little config. Not a tradeoff I wish we had to make, but considering everything else works, I think it's fine.

IMax153: Yea - you're right, I just get annoyed by stuff like this 🤣 .

IMax153 (approved):
@aniravi24 - thanks for the work on this! If you give the 👍 this looks good to me.

aniravi24:
> @aniravi24 - thanks for the work on this! If you give the +1 this looks good to me and I can merge.

LGTM!

PR #50, opened 2023-04-22
docs(style-guide): writing code with effect

aniravi24:


IMax153:
@aniravi24 - I think this content may be better suited for the "Style Guide" given that we're not really _organizing_ Effects, per se, but instead talking about different ways to write code with Effect.

Perhaps we can move this content to the "Style Guide" under a heading called something like "Writing Code with Effect". Then you can have three subheadings: "Using the Pipeable API", "Using Generators", "Using the Do Simulation" (names are up for debate). And then, under the pipeable section you can discuss the fact that most non-constructor methods are dual in nature (i.e. can be data-first or data-last).

aniravi24:
> @aniravi24 - I think this content may be better suited for the "Style Guide" given that we're not really _organizing_ Effects, per se, but instead talking about different ways to write code with Effect.
> 
> Perhaps we can move this content to the "Style Guide" under a heading called something like "Writing Code with Effect". Then you can have three subheadings: "Using the Pipeable API", "Using Generators", "Using the Do Simulation" (names are up for debate). And then, under the pipeable section you can discuss the fact that most non-constructor methods are dual in nature (i.e. can be data-first or data-last).

I was debating whether to put this in the style guide or make it a separate page - I will go with your suggestion for now.

pages/docs/style-guide.mdx
```diff
@@ -30,6 +32,109 @@ There is only one exception. Functions from module Function shall be used direct
...
+  )
+)
+```
+
+Here, Random.next() is getting piped as an argument to `Effect.flatMap`.
+
+<Callout>
+  Many functions in Effect have `dual` APIs, which means that you can either
+  call it `data-first` or `data-last`. Here, you can see how `Effect.flatMap`
+  was already defined, so `Random.next()` is passed in last. Similarly, you can
```
IMax153, line 69:
```suggestion
  is _after_ the call to `Random.next()` in the `pipe`, so `Random.next()` is passed in last (i.e. `data-last`). Similarly, you can
```

pages/docs/style-guide.mdx
```diff
@@ -30,6 +32,109 @@ There is only one exception. Functions from module Function shall be used direct
...
+  also call `Effect.flatMap(Random.next(), (n) => ...)` with `Random.next()` and
+  without `pipe` to get the same result.
```
IMax153, line 71:
```suggestion
  also call `Effect.flatMap(Random.next(), (n) => ...)` directly without `pipe` with `Random.next()` as the first argument (i.e. `data-first`) get the same result.
```

pages/docs/style-guide.mdx
```diff
@@ -30,6 +32,109 @@ There is only one exception. Functions from module Function shall be used direct
...
+  const n = yield* $(Random.next())
+  if (n > 0.5) {
+    return yield* $(Effect.succeed("yay!"))
+  } else {
+    return yield* $(Effect.fail("oh no!"))
+  }
+})
+```
+
+Note all the asterisks and dollar signs. The `*` is used to denote a generator, and the `$` function which is passed in as an argument to `Effect.gen` must follow `yield*` if you are yielding an Effect.
```
IMax153, line 92:
```suggestion
Note the `yield*`, which is used to yield the next value from the generator. Additionally, one can also observe the `$` argument passed to `Effect.gen`, which is just a function that serves as an "adapter" that allows TypeScript to track our Effect types. When yielding an Effect, you must pass the Effect you would like to `yield*` to the adapter function to ensure that TypeScript can properly infer the types.
```

pages/docs/style-guide.mdx
```diff
@@ -30,6 +32,109 @@ There is only one exception. Functions from module Function shall be used direct
...
+
+This way, you can write in a more imperative style rather than having to use `pipe` and `flatMap`.
+
+<Callout type="info">
+  The `$` argument can actually also be used as a `pipe`. For example, you could
```
IMax153, line 97: Two suggestions:
1. I would note that you can call the `$` adapter function whatever you want, it doesn't have to be called `$`
2. I would also maybe show a quick example of using `$` as a `pipe`

pages/docs/style-guide.mdx
```diff
@@ -30,6 +32,109 @@ There is only one exception. Functions from module Function shall be used direct
...
+```ts twoslash
+import * as Effect from "@effect/io/Effect"
+import * as Random from "@effect/io/Random"
+import { pipe } from "@effect/data/Function"
+
+const example = pipe(
+  Effect.Do(),
+  Effect.let("randomValue", () => 3),
+  Effect.bind("randomNumber", ({ randomValue }) =>
+    randomValue > 2 ? Random.next() : Random.next()
```
IMax153, line 121: Is there a reason we call `Random.next()` twice here?

IMax153 (changes requested):

pages/docs/style-guide.mdx
aniravi24: I was doing another read at my original copy and definitely wanted to change it, thanks for these changes! Will do another pass.

pages/docs/style-guide.mdx
```diff
@@ -30,6 +32,109 @@ There is only one exception. Functions from module Function shall be used direct
...
+```ts twoslash
+import { pipe } from "@effect/data/Function"
+import * as Effect from "@effect/io/Effect"
+import * as Random from "@effect/io/Random"
+
+const example = pipe(
+  Random.next(),
+  Effect.flatMap((n) =>
+    n > 0.5 ? Effect.succeed("yay!") : Effect.fail("oh no!")
+  )
```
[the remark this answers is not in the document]
aniravi24: I think prettier will respect newlines but you can try running lint to see!

pages/docs/style-guide.mdx
aniravi24: I just wanted to choose some Effect I could call but it does look confusing. Mainly trying to show how `randomValue` is now accessible from that line

IMax153:
@aniravi24 - just ping me again for a review once all the comments are addressed and you've got CI passing (looks like just a `pre-commit` issue).

aniravi24:
> @aniravi24 - just ping me again for a review once all the comments are addressed and you've got CI passing (looks like just a `pre-commit` issue).

Ready!

alex-dixon:
Just read through and thought it looks good :)

IMax153:
Given #59, I'm going to ping @mikearnaldi here for a final say.

PR #51, opened 2023-04-22
docs(core-concepts): add some more content to error handling docs

aniravi24:


pages/docs/core-concepts/error-handling.mdx
```diff
@@ -25,24 +25,28 @@ Expected errors, also known as _failures_, _typed errors_ or _recoverable errors
 To indicate that our program can fail for some expected reason, we can use the `Effect.fail` constructor:
 
 ```ts twoslash
-import * as Effect from "@effect/io/Effect"
+import * as Effect from "@effect/io/Effect";
```
IMax153, line 28: Please remove semi-colons from the code samples - it's just extra syntax noise.

pages/docs/core-concepts/error-handling.mdx
```diff
@@ -139,14 +144,70 @@ We could optionally also use the `Effect.catchTags` combinator to handle all err
...
   Effect.catchTags({
     FooError: (fooError) => Effect.succeed(`Recovering from ${fooError._tag}`),
-    BarError: (barError) => Effect.succeed(`Recovering from ${barError._tag}`)
+    BarError: (barError) => Effect.succeed(`Recovering from ${barError._tag}`),
   })
-)
+);
+```
+
+#### Modifying the error channel
```
IMax153, line 157: We're not really "modifying" the error channel, we're mapping over the error channel.

pages/docs/core-concepts/error-handling.mdx
```diff
@@ -139,14 +144,70 @@ We could optionally also use the `Effect.catchTags` combinator to handle all err
...
+    }
+  })
+);
+```
+
+`Effect.mapError` is similar to `Effect.map`, except instead of operating on the success channel, it operates on the error channel.
+
+#### Example: Handling errors from Promises
+
+If you are using Effect with a Promise-based API, you can use the `Effect.promise` constructor to convert a Promise into an Effect. If the Promise rejects, the resulting Effect will throw with the error that the Promise rejected with. You can also use other variants such as `Effect.tryPromise` or `Effect.tryCatchPromise` to convert Promise rejections into typed errors in the Effect error channel.
```
IMax153, line 185: There are a few issues with this paragraph that I feel deserve pointing out - `promise` doesn't throw, it fails with a defect. `tryPromise` doesn't convert to a typed error, it elects failures to recoverable errors with type `unknown` (i.e. it is up to the user to refine the type). `tryCatchPromise` is the only one that elects failures to typed errors.

I would frame these three methods a bit differently:
1. `promise` - any errors thrown by the `Promise` provided to this constructor will result in a defect (i.e. an unrecoverable error)
2. `tryPromise` - any errors thrown by the `Promise` provided to this constructor will result in a recoverable error with type `unknown` in the error channel. It is then left to the user to refine the type further.
3. `tryCatchPromise` - any errors thrown by the `Promise` provided to this constructor will be provided to the `onReject` function, allowing users to elect `Promise` rejects to typed errors in the error channel

IMax153 (changes requested):

pages/docs/core-concepts/error-handling.mdx
aniravi24: Sure, pending resolution to https://github.com/Effect-TS/website/pull/54

aniravi24: Will re-format once the build config PR goes in

pages/docs/core-concepts/error-handling.mdx
aniravi24: Sorry for the confusion here, was trying to simplify the language a bit and it may have been misleading.

pages/docs/core-concepts/error-handling.mdx
aniravi24: Thanks for the feedback! Made some tweaks to reflect this information - might have simplified a bit too much initially.

pages/docs/core-concepts/error-handling.mdx
```diff
@@ -152,6 +154,62 @@ const program = pipe(
...
+Sometimes you don't want to actually handle the error, but map over it instead. For example, you might want to add some additional information to the error, or you might want to change the error type altogether.
+
+To do this, we can use `Effect.mapError`:
+
+```ts twoslash
+// @include: errors-1
+// ---cut---
+
+const program = pipe(
+  //    ^?
```
IMax153, line 168:
```suggestion
const program = pipe(
//    ^?
```

pages/docs/core-concepts/error-handling.mdx
```diff
@@ -152,6 +154,62 @@ const program = pipe(
...
+const promise = Promise.resolve(new Error("oh no!"))
+
+const effectPromise = Effect.promise(() => promise)
+//    ^?
+
+const effectTryPromise = Effect.tryPromise(() => promise)
+//    ^?
+
+const effectTryCatchPromise = Effect.tryCatchPromise(
+  //    ^?
```
IMax153, line 207:
```suggestion
const effectTryCatchPromise = Effect.tryCatchPromise(
//    ^?
```

pages/docs/core-concepts/error-handling.mdx
```diff
@@ -152,6 +154,62 @@ const program = pipe(
 )
 ```
 
+#### Mapping over the error channel
```
IMax153, line 157:
```suggestion
#### Manipulating Expected Errors
```
Was just trying to think of a way to keep this heading relatively consistent with the above heading (i.e. `Catching Expected Errors`).

IMax153 (changes requested):
@aniravi24 - I only requested some minor changes. The content looks great though - will be an excellent addition to the docs!

I've also been thinking that I should expand upon the `Catching Expected Errors` section to include documentation for the other `catch*` combinators.

pages/docs/core-concepts/error-handling.mdx
aniravi24: Mine had said "Modifying", but you were right in that you're not technically modifying it, although I had simplified it that way because the error channel changes from that point forward to the new value you mapped it to.

pages/docs/core-concepts/error-handling.mdx
aniravi24: Prettier will auto format this comment back to where it was - it aligns the comment to the code underneath it as far as I can tell.

pages/docs/core-concepts/error-handling.mdx
aniravi24: Same as above

aniravi24:
> @aniravi24 - I only requested some minor changes. The content looks great though - will be an excellent addition to the docs!
> 
> I've also been thinking that I should expand upon the `Catching Expected Errors` section to include documentation for the other `catch*` combinators.

Covering some of the others may require understanding what defects are first, so maybe we do that section first and then see how to fit in the other catch combinators?

IMax153:
I was rather thinking that we can cover the catch combinators that have to do with expected errors here, and cover the other catch combinators under defects. Something like "Catching Unexpected Errors" later on.

pages/docs/core-concepts/error-handling.mdx
IMax153: Alright I think we can just go with "Manipulating Expected Errors" for now if you're cool with that - can always change it later, but at least this way the headings are consistent.

aniravi24: Sure

IMax153 (approved):
Nice work @aniravi24 ! LGTM 👍

PR #54, opened 2023-04-23
some build config improvements

aniravi24:


aniravi24:
@IMax153 can we add prettier to the project? Very low config and would make it really easy to format everything cause it "just works". I find myself regularly wanting to format pages and force format them with prettier, which causes a lot of other lines to get changed.

IMax153:
> @IMax153 can I add prettier to the project? 

To avoid having too many tools between our repositories, I would prefer to integrate our own eslint plugin for `dprint` over `prettier` like we do in all the other repositories. If it's not possible, let me know and we can discuss.

Here's an example of how we integrate it in other projects: https://github.com/Effect-TS/io/blob/main/.eslintrc.cjs#L64. Though we now publish this package through the `effect` org itself, so instead of `@repo-tooling` it would be `@effect` - see https://www.npmjs.com/package/@effect/eslint-plugin.

IMax153:
@aniravi24 - I just did some testing locally - I don't think our current setup with `dprint` will work, so feel free to introduce a `prettier` / `eslint` setup that works into the repository.

aniravi24:
> @aniravi24 - I just did some testing locally - I don't think our current setup with `dprint` will work, so feel free to introduce a `prettier` / `eslint` setup that works into the repository.

I've got something working now.

A couple notes though:

1. Prettier mangles mermaid formatting so you do have to be careful to use a `prettier-ignore` comment as it doesn't support that yet
2. I kept formatting separate from eslint here (usually because running it through eslint makes it slower and more memory hungry), but since this is a docs site and not a full blown codebase, let me know if you'd prefer that it's directly integrated into the linting step for convenience.
3. It automatically adds a bit of spacing to comments, which can affect the `^?`. But I think if there's a proper formatting step used correctly in the editor this being incorrectly positioned is easily avoided. The start of the comment doesn't appear to matter, it's just where the `^?` shows up.

Let me know what you think @IMax153

IMax153:
@aniravi24 - this looks great so far! Thanks for the work you've put in. Here are my thoughts on your questions:

1. I think this is fine, we probably won't have too many mermaid formatted things, but it's good to know regardless. Maybe we can drop something in the `README` to this effect.
2. I think this is fine as an initial step, but if we are going to ignore formatting `.mdx` code blocks, perhaps we should consider re-using the ESLint integration we have setup with `dprint` in all the other repos? In our experience it's been much faster than `prettier` and since we are not going to worry about the `*.mdx` stuff maybe it's less complex to just keep things the same as in our other repos? Let me know what you think.
3. This is totally fine - we can make adjustments to the existing code to add spacing where necessary.

aniravi24:
> @aniravi24 - this looks great so far! Thanks for the work you've put in. Here are my thoughts on your questions:
> 
> 1. I think this is fine, we probably won't have too many mermaid formatted things, but it's good to know regardless. Maybe we can drop something in the `README` to this effect.
> 2. I think this is fine as an initial step, but if we are going to ignore formatting `.mdx` code blocks, perhaps we should consider re-using the ESLint integration we have setup with `dprint` in all the other repos? In our experience it's been much faster than `prettier` and since we are not going to worry about the `*.mdx` stuff maybe it's less complex to just keep things the same as in our other repos? Let me know what you think.
> 3. This is totally fine - we can make adjustments to the existing code to add spacing where necessary.

Sorry, I think I was unclear with the second point - Prettier will format everything, I just meant that integrating into ESLint wasn't as straightforward as using the prettier plugin for ESLint as ESLint doesn't read MDX by default. Running prettier separately without ESLint formats everything just fine. Would you prefer that I try to get eslint parsing MDX files as well or is prettier running on its own fine too?

IMax153:
> Would you prefer that I try to get ESLint parsing MDX files as well

@aniravi24 - if we can do this it might be a good thing. This way we can have consistent formatting in the MDX files as well. Especially because these `*.mdx` files will likely make up a large part of the repo.

aniravi24:
> > Would you prefer that I try to get ESLint parsing MDX files as well
> 
> @aniravi24 - if we can do this it might be a good thing. This way we can have consistent formatting **and** linting in the MDX files as well. Especially because these `*.mdx` files will likely make up a large part of the repo.

Got it to work with eslint! Not sure what the `next lint` command does, but couldn't get it to work when passing multiple extensions with `--ext` so using eslint directly instead.

pages/docs/core-concepts/error-handling.mdx
```diff
@@ -104,7 +107,7 @@ Given that `FooError` has a `_tag` field, we can use the built-in `catchTag` met
 // @include: errors-1
 // ---cut---
 const program = pipe(
-//    ^?
+  //    ^?
```
IMax153, line 110: Can we remove the leading spaces for all of these instances? Just to make sure we keep things consistent.

IMax153 (approved):
@aniravi24 - These changes look great! I tested on my machine as well and it looks like things are working as expected! This gives us a fantastic framework to add in linting rules over time, so thank you!

Once the merge conflict is addressed (sorry that was my fault 😅 ), we can get this in!

pages/docs/core-concepts/error-handling.mdx
aniravi24: Unfortunately, this is dependent on prettier, I'm not sure if there's a way to do that. It's aligning the comment with the code underneath it since it's on the next line. Doesn't seem like there's a way I can prevent it from doing that.

theme.config.tsx
```diff
@@ -1,8 +1,9 @@
 import { useRouter } from "next/router";
 import { EffectLogo } from "./components/EffectLogo";
+import { DocsThemeConfig } from "nextra-theme-docs";
 
 // eslint-disable-next-line import/no-anonymous-default-export
```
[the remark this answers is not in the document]
aniravi24: Yup!

PR #60, opened 2023-04-30
Add Cookbook page

datner:
I'm on the fence if this should exist or not, but it could be a good place to see how people use effect and help people build intuitions

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,8 @@
   "devDependencies": {
     "@effect/data": "^0.12.2",
     "@effect/io": "^0.25.7",
+    "@effect/match": "^0.18.1",
+    "@effect/schema": "^0.17.2",
     "@mdx-js/mdx": "^2.3.0",
     "@types/node": "^18.16.1",
     "@types/react": "^18.2.0",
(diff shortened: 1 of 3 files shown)


aniravi24:
I don't think this should live inside the docs - I suspect it will get messy and hard to maintain really fast, given the size of the ecosystem of libraries growing, and the feature sets of these libraries.

I'm not sure the best way to maintain or organize something like this, but I'd probably put it in a separate repo.

datner:
@aniravi24 
I generally agree, but as a community member said:

> It still might be useful for some random things. There is a good snippet here for using effect/schema with zod libs, for example, that I know I'm going to have to spend 20min searching for again at some point in the future.

So I made this PR and delegated the decision on the matter to the powers at be lol

mikearnaldi:
I am mixed, I do think this type of content has a place in the website but not in a prominent manner, probably a section "Advanced & Crazy Usage" something targeting the curious

datner:
@mikearnaldi So what's our next step?

datner:
bump

PR #63, opened 2023-05-05
Update services.mdx

samuelebonini:
docs mention a `provide` function; the actual name is `provideService`


IMax153:
Thanks @samuelebonini !

PR #88, opened 2023-06-21
add Using Generators

gcanti:


IMax153:
@gcanti - if we're going to have a page on using generators under Effect essentials, do we still need it under idiomatic code?

IMax153:
Also it may be worth noting somehow that there is a small amount of runtime overhead when using generators in JS? Not sure.

gcanti:
@IMax153 I think we can remove the entire "Writing Code with Effect" section and move the part about do notation to a separate page

IMax153:
> I think we can remove the entire "Writing Code with Effect" section and move the part about do notation to a separate page

This makes sense to me! 👍

PR #110, opened 2023-06-23
add Immutability

gcanti:


mikearnaldi:
Closing all PRs targeting old, we can directly port it in `website-redesign` if we believe it is important

PR #132, opened 2023-06-29
Add blog to website

IMax153:


pages/blog/type-safe-errors/naive-approach.mdx
```diff
@@ -0,0 +1,163 @@
...
+description: 'In this article we will take a look at the usual way we represent errors in TypeScript'
+date: June 7th, 2023
+---
+
+# A naive approach using classes with instanceof
+
+The first step to achieve type safety in error management as one may imagine is to define errors, let's have a look at a usual piece of code with the objective of improving its safety.
+
+```ts
+import * as S from "@effect/schema/Schema"
```
IMax153, line 11: **Suggestion**: 

Given that we explicitly [recommend fully qualified name-spaced imports in the documentation](https://effect.website/docs/guide/style/modules), I would suggest being explicit in the blog as well. I.E:

```ts
import * as Schema from "@effect/schema/Schema"
```

PR #164, opened 2023-07-07
update 'Effect Essentials' to new api

ethanniser:
this pr:

- updates all outdate code and explanations in 'Effect Essentials' to match the [new api](https://github.com/Effect-TS/io/pull/461)
- modified examples where `random.next`, `nextBoolean` and `nextInt` were called as methods when they are just values

to ensure everything is correct [here is a typescript playground with every single snippet from this section with no errors](https://tsplay.dev/m07xRN) (give it a sec to load the imported packages)

also

I was unable to start my dev server due to some [nasty (webpack?) errors ](https://pastebin.com/0rQzKJP1), so I dont know exactly how things will look but the changes are pretty minor so I assume everything should be fine

### DIFF
--- a/pages/docs/essentials/control-flow.mdx
+++ b/pages/docs/essentials/control-flow.mdx
@@ -65,9 +65,7 @@ import * as Effect from "@effect/io/Effect"
 import * as Random from "@effect/io/Random"
 
 // Effect<never, never, Option.Option<number>>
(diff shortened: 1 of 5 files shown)


gcanti:
Thank you @ethanniser, going to merge and fix a few nitpicks later

PR #186, opened 2023-07-14
Add  entry in the creating Effect cheatsheet

codella:
While reading the [Creating Effects](https://effect.website/docs/essentials/creating) page on `effect.website`, I noticed that `Effect.promise` was missing in the [Cheatsheet](https://effect.website/docs/essentials/creating#cheatsheet).

This PR aims to fix that. Below an image of how it appears locally, running `pnpm dev`.

Fri Jul 14 10:15:27 PM CEST 2023


gcanti:
good catch, thanks

PR #188, opened 2023-07-15
Make flattening explicit in Pipeline page's sheatsheet

codella:


gcanti:
@codella I'd keep the cheatsheet as simple as possible, being in "Essentials" this page will be read by new users and I wouldn't overload them with too much info at this point (the error and context tracking is explained later on)

codella:
@gcanti makes sense 👍🏼 I wasn't really sure whether or not propose this change.

PR #211, opened 2023-07-19
Update `try.ts` example to use the `try` function

korokd:
The code example for the simple use case of the `try` function uses the `sync` function. This fixes that.

### DIFF
--- a/src/essentials/creating/try.ts
+++ b/src/essentials/creating/try.ts
@@ -1,7 +1,7 @@
 import { Effect } from "effect"
 
 // $ExpectType Effect<never, never, number>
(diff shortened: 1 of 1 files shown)


gcanti:
@korokd I think I just copied and pasted from the `sync` example, the code is completely wrong, should be

```ts

// Effect<never, unknown, any>
const program = Effect.try(
  () => JSON.parse("") // JSON.parse may throw for bad input
)
```

korokd:
@gcanti I added a commit to this PR with the code you typed in case you wanna just re-use this PR

gcanti:
@korokd perfect, thank you!

PR #241, opened 2023-08-03
fix Counter.ts in Ref guide

erikshestopal:
Fixes the following issue running the code sample: 

```ts
node:internal/process/esm_loader:46
      internalBinding('errors').triggerUncaughtException(
                                ^

Error [TypeError]: Cannot read properties of undefined (reading 'modify')
TypeError: Cannot read properties of undefined (reading 'modify')
    at <anonymous> (/Users/erikshestopal/effect-ts-playground/node_modules/effect/node_modules/@effect/io/src/internal/ref.ts:119:52)
```

### DIFF
--- a/src/guide/state-management/Counter.ts
+++ b/src/guide/state-management/Counter.ts
@@ -1,18 +1,14 @@
 import { Effect, Ref } from "effect"
 
 export class Counter {
-  constructor(private value: Ref.Ref<number>) {}
+  inc: Effect.Effect<never, never, void>
+  dec: Effect.Effect<never, never, void>
+  get: Effect.Effect<never, never, number>
 
-  get inc() {
-    return Ref.update(this.value, (n) => n + 1)
-  }
-
-  get dec() {
-    return Ref.update(this.value, (n) => n - 1)
-  }
-
-  get get() {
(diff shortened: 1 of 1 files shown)


gcanti:
Thanks @erikshestopal, could you please provide me with a repro that I can run? This will help me to track down the problem

gcanti:
> If you run the previous code using the examples on https://effect.website/docs/state-management, you'll see that it's invalid code

Actually I don't, that's because of how typescript compiles the code I guess:

```js
import { Effect, Ref } from 'effect';
export class Counter {
    constructor(value) {
        this.value = value;
        this.inc = Ref.update(this.value, (n) => n + 1);
        this.dec = Ref.update(this.value, (n) => n - 1);
        this.get = Ref.get(this.value);
    }
}
```

> Another way to fix this code is to initialize the properties in the constructor:

Yeah, seems like a good solution to me

gcanti:
@erikshestopal yes, that's what I did to reproduce the error. I didn't realize that I had written incorrect code because I didn't see any errors while running the snippets with ts-node. The code needs to be corrected, and of the two suggestions, the second one seems good to me.

gcanti:
Thank you!

PR #254, opened 2023-08-18
Add Fibers page, move Interruption Model

evelant:


pages/docs/concurrency/fibers.mdx
```diff
@@ -0,0 +1,55 @@
...
+A fiber is created any time an effect is run. When running effects concurrently, a fiber is created for each concurrent effect.
+
+## When do Fibers run?
+
+New fibers begin execution after the current fiber completes or yields. This is necessary to prevent infinite loops in some cases, and is useful to know when using the `fork` APIs.
+
+In the following example the `SubscriptionRef` `changes` stream only contains a single value `2` because the fiber (created by `fork`) to run the stream is started _after_ the value has been updated.
+
+```ts
+Effect.gen(function* (_) {
```
gcanti, line 27: Could you please move the examples to the `src` folder (as files)? Then you can import the code adding:

```
file=<rootDir>/src/path/to/my/example.ts
```

to the ts fence. 

See for example https://github.com/Effect-TS/website/blob/dce5bbd43f0a580a94a873b43ea5d74228e54102/pages/docs/concurrency/concurrency-options.mdx#L32

This will allow us to run `pnpm check` to ensure all examples compile.

pages/docs/_meta.json
```diff
@@ -12,7 +12,6 @@
   },
   "essentials": "Effect Essentials",
   "error-management": "Error Management",
-  "interruption-model": "Interruption Model",
```
gcanti, line 15: If we move the `interruption-model.mdx` file there are a couple of links in the following files to fix:

- `unexpected-errors.mdx`
- `logging.mdx`

(search for `"interruption-model"` in `./pages/docs`)

gcanti (changes requested):

pages/docs/_meta.json
evelant: Fixed the links, thanks for the feedback, sorry I missed that.

gcanti (approved):

PR #268, opened 2023-08-30
fixed the line specification because it was incorrect

mrsekut:
Line references were out of alignment and have been corrected.

### before
image

### after
image


gcanti:
Thanks @mrsekut

PR #320, opened 2023-09-26
Update docs to reflect Stream.contramap rename to mapInput

GraemeF:


gcanti:
Thanks @GraemeF , we should also rename the `src/streaming/sink/operations/contramap.ts` file

PR #363, opened 2023-11-02
Add other installation methods

DenisGorbachev:


pages/docs/essentials/importing.mdx
```diff
@@ -11,10 +12,30 @@ in most cases, installing the `effect` package is all you need to get started. L
...
+
 ```bash filename="Terminal"
 npm install effect
 ```
 
+</Tab>
+<Tab>
+
+```bash filename="Terminal"
+pnpm install effect
```
gcanti, line 26: shouldn't here be `pnpm add effect`?

from `pnpm -h`:

|            |                                                                                                                       |
| ---------- | --------------------------------------------------------------------------------------------------------------------- |
| add        | Installs a package and any packages that it depends on. By default, any new package is installed as a prod dependency |
| i, install | Install all dependencies for a project                                                                                |

gcanti (changes requested):

pages/docs/essentials/importing.mdx
DenisGorbachev: Fixed 👍 

Sorry, I read that pnpm was a drop-in replacement for npm & assumed the equality of subcommand names.

gcanti (approved):

PR #366, opened 2023-11-04
Add some more information about `flatMap`

IMax153:


src/faq/array/flattening-array.ts
```diff
@@ -0,0 +1,7 @@
+import { pipe, Effect } from "effect"
+
+// Effect.Effect<never, never, Array<number>>
```
gcanti, line 3: should `// $ExpectType Effect<never, never, number[]>`

src/faq/array/flattening-readonlyarray.ts
```diff
@@ -0,0 +1,7 @@
+import { pipe, Effect, ReadonlyArray } from "effect"
+
+// Effect.Effect<never, never, ReadonlyArray<number>>
```
gcanti, line 3: idem, should be `// $ExpectType Effect<never, never, number[]>`

src/faq/array/flattening-array.ts
gcanti: I would also move these two files to `/src/essentials/pipeline`

pages/docs/faq.mdx
```diff
@@ -41,6 +41,32 @@ In practice, it's important to note that you should typically only run Effects a
...
+### A Closer Look at `flatMap`
+
+Many JavaScript / TypeScript engineers are familiar with the term `flatMap` due to it's [presence as a method on the `Array` prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap). Therefore it may be confusing to see `flatMap` methods exported from many of the modules that Effect provides.
+
+The `flatMap` operation can actually be used to describe a more generic data transformation. Let's consider for a moment a generic data type that we will call `F`. `F` will also be a container for elements, so we can further refine our representation of `F` to `F<A>` which states that `F` holds some information about some data of type `A`.
+
+If we have an `F<A>` and we want to transform to an `F<B>`, we could use the `map` operation:
+
+```
+map: F<A> -> (A -> B) -> F<B>
```
gcanti, line 53: This looks too much Haskell-y IMO, I would use a more TypeScript-y signature here, like
```
map: <A, B>(fa: F<A>, f: (a: A) => B) => F<B>
```

pages/docs/faq.mdx
```diff
@@ -41,6 +41,32 @@ In practice, it's important to note that you should typically only run Effects a
...
+```
+
+But what if we have some function that returns an `F<B>` instead of just `B`? We can't use `map` because we would end up with a `F<F<B>>` instead of an `F<B>`. What we really want is some operator that allows us to `map` the data and then `flatten` the result.
+
+This exact situation describes a `flatMap` operation:
+
+```
+flatMap: F<A> -> (A -> F<B>) -> F<B>
```
gcanti, line 61: and then
```
flatMap: <A, B>(fa: F<A>, f: (a: A) => F<B>) => F<B>
```

pages/docs/faq.mdx
```diff
@@ -41,6 +41,32 @@ In practice, it's important to note that you should typically only run Effects a
...
+```
+
+You can also see how this directly applies to `Array`'s `flatMap` if we replace our generic data type `F` with the concrete data type of `Array`:
+
+```
+flatMap: Array<A> -> (A -> Array<B>) -> Array<B>
```
gcanti, line 67:
```
flatMap: <A, B>(fa: Array<A>, f: (a: A) => Array<B>) => Array<B>
```

pages/docs/essentials/pipeline.mdx
```diff
@@ -121,6 +121,35 @@ If you simply want to replace the value with a constant, you can use `Effect.as`
...
+  <details>
+  <summary>Flattening a Nested Array</summary>
+
+  If you have an Effect where result type is a nested `Array`, for example:
+
+  ```
+  Effect<never, never, Array<Array<A>>>
+  ```
+
+  you can easily flatten the array using either the `ReadonlyArray` module exported from Effect:
```
gcanti, line 138: what about

```diff
-exported from Effect:
+exported from `effect`:
```

gcanti (changes requested):

src/faq/array/flattening-array.ts
IMax153: Done 👍

src/faq/array/flattening-readonlyarray.ts
IMax153: Done 👍

pages/docs/faq.mdx
IMax153: Done 👍

pages/docs/faq.mdx
IMax153: Done 👍

pages/docs/faq.mdx
IMax153: Done 👍

pages/docs/essentials/pipeline.mdx
IMax153: Done 👍

gcanti (approved):

gcanti:
Thanks @IMax153

PR #371, opened 2023-11-09
Runtime: fix top-level-runtime-configuration example

gcanti:


src/runtime/toRuntime.ts
```diff
@@ -0,0 +1,30 @@
...
+const appLayer = Logger.replace(
+  Logger.defaultLogger,
+  Logger.make(({ message }) => console.log(message))
+)
+
+// Create a scope for resources used in the configuration layer
+const scope = Effect.runSync(Scope.make())
+
+// Transform the configuration layer into a runtime
+const runtime = Effect.runSync(
```
tim-smart, line 13: Sometimes layers require async construction. This will raise an error in that case - something to take note of.

src/runtime/toRuntime.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// Execute the program using the custom runtime
+runSync(program)
+/*
+Output:
+Application started!
+*/
+
+// Cleaning up any resources used by the configuration layer
+Effect.runSync(Scope.close(scope, Exit.unit))
```
tim-smart, line 30: Similar to above, you can have async finializers, so runFork is a better option here.

PR #380, opened 2023-12-01
Update effect version

KhraksMamtsov:
Can't figure out what to do with batching example - need help)

- move from `.use` to `.provide`
- remove `.use` mentions


gcanti:
Thank you @KhraksMamtsov, going to merge this and then deal with the batching example in another PR

PR #493, opened 2024-02-29
Add playground

tim-smart:


src/CodeEditor/rx/workspace.ts
```diff
@@ -84,8 +84,9 @@ export const workspaceHandleRx = Rx.family((workspace: Workspace) =>
           })
         )
 
-        yield* get.stream(solved).pipe(
-          Stream.drop(1),
+        yield* get.stream(solved, {
+          withoutInitialValue: true
```
tim-smart, line 88: 👏 Sometimes I forget the APIs in my own libraries.

PR #545, opened 2024-04-04
add missing comma

ethanniser:
adds missing comma, otherwise example is invalid syntax


tim-smart (approved):
Who even wrote that code? Geez.

PR #577, opened 2024-04-23
UI  - Fix Search Input autofocus on Safari

suddenlyGiovanni:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Bug:
On Safari, the "cmd + k" keyboard shortcut opens the Search Dialog, but the input field doesn't get focused as expected. Other Chromium-based browsers do not suffer the same issue.

https://github.com/Effect-TS/website/assets/15946771/2d070737-026c-4b3b-90e4-8c5bc1edda44

 
### Issue:
The query selector used to grab the search box element from the DOM had a `setTimeout` set to **1ms**.

### Fix:  
Tentatively increased said `setTimeout` to **100ms** . This does the trick.
A possible alternative solution could be to use the `React.useRef` API instead on the `SearchBox` component.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/src/components/atoms/search.tsx
+++ b/src/components/atoms/search.tsx
@@ -39,7 +39,7 @@ export const Search: FC<{ className?: string }> = ({ className = "" }) => {
       setTimeout(
         //@ts-ignore
         () => document.querySelector(".ais-SearchBox-input")?.focus(),
-        1
+        100
       )
     } else {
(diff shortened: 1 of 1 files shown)


suddenlyGiovanni:
thx, now on prod it works like a charm 🌟

PR #579, opened 2024-04-24
Remove usage of `EffectAdapter`

fubhy:


fubhy:
@mikearnaldi should we merge this yet or wait for it to prove itself in the wild first?

mikearnaldi:
Let's go ahead and merge, I am now confident enough that TS 5.5 will have the fix

riwsky:
@mikearnaldi what's the fix in question here? Is adapterless Effect.gen use somehow incorrect on TS versions <4.5?

fubhy:
> @mikearnaldi what's the fix in question here? Is adapterless Effect.gen use somehow incorrect on TS versions <4.5?

There was a bug in TypeScript v5.4 and below that was fixed in TypeScript v5.5. It used to lead to type inference issues in some scenarios.

riwsky:
How would the bug manifest—rejecting acceptable programs? or allowing unacceptable programs? I'm using adapterless .gen in a 5.4 codebase right now, and not hitting errors.

Is the issue in question https://github.com/microsoft/TypeScript/issues/57690 ?

mikearnaldi:
The bug was: https://github.com/microsoft/TypeScript/pull/58337

riwsky:
@mikearnaldi word. thanks!

PR #580, opened 2024-04-24
Rename `ReadonlyArray` to `Array` module in docs.

fubhy:


gcanti:
@fubhy but why didn't we receive any errors before these changes?

fubhy:
@gcanti I don't know how the build process works for the examples in this repo and whether they are checked fully? Are they? If yes, then maybe because the `effect` version was still on 2.x.x`?

fubhy:
If they are not checked, btw. I think we should add that ... Kinda important to have all docs (and README's, btw.) checked for compatibility with the latest version.

PR #599, opened 2024-04-29
use different .retry options for complexity examples

tim-smart:


tim-smart:
I think the code is better, but it for some reason breaks the syntax highlighting.

PR #612, opened 2024-05-07
Fix documentation typo

arthurvmdantas:
Namespace imports (and not named imports) have a harder tree shakeability.

## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [X] Documentation Update

## Description

Just fixing a documentation typo.

## Related

- Related Issue #
- Closes #

### DIFF
--- a/content/docs/400-guides/100-essentials/100-importing-effect.mdx
+++ b/content/docs/400-guides/100-essentials/100-importing-effect.mdx
@@ -72,7 +72,7 @@ import * as Effect from "effect/Effect"
 Both forms of import allow you to access the functionalities provided by the `Effect` module.
 
 However an important consideration is tree shaking, which refers to a process that eliminates unused code during the bundling of your application.
-Named imports may generate tree shaking issues when a bundler doesn't support deep scope analysis.
+Namespace imports may generate tree shaking issues when a bundler doesn't support deep scope analysis.
 
 Here are some bundlers that support deep scope analysis and thus don't have issues with named imports:
 


datner:
We're talking about named imports:
```ts
```
vs namespace imports
```ts
import * as Effect from "effect/Effect"
```
You're thinking about named imports:
```ts
import { map } from "effect/Effect"
```
vs namespace imports I believe.

We consider `import { map } from "effect/Effect"` to be extremely rare usage of effect, and the other two forms as idiomatic with one being more comfortable and the other easier to tree-shake.

If you think something in our understand is off, please correct me

arthurvmdantas:
Hello, @datner 

The documentation says:

> Both forms of import allow you to access the functionalities provided by the Effect module.

> However an important consideration is tree shaking, which refers to a process that eliminates unused code during the bundling of your application. **Named imports may generate tree shaking issues when a bundler doesn't support deep scope analysis**.

I'm sorry if I'm being a little bit "dumb" here as I'm still learning JS/TS, but why would the following import (a named import) cause a tree shaking issue? The code below is the only example of named imports on the page. Named imports are easier for bundlers to process, right?

```js
import { Effect } from "effect" // named import
```

Don't you think then that the documentation meant to say "_namespace imports may generate tree shaking issues when..._" instead?

```js
import * as Effect from "effect/Effect" // namespace import
```

datner:
@arthurvmdantas Not in this case, no 😅  `Effect` from `"effect"` is a named import, but much more importantly it's a namespace re-export!
```ts
/// @index.ts
export * as Effect from "effect/Effect"
// ...
```

Most bundlers know how to correctly resolve and treeshake namespace imports, not as good as regular named exports, but esbuild-like bundlers don't treeshake `export * as Module from "Module"` that is subsequently named-imported _at all_

meaning that
```ts
```
might import the entire `Effect` module in some bundlers.
But
```ts
import * as Effect from "effect/Effect"
```
will treeshake almost entirely.

You can also see [other sufficiently-large libs](https://mui.com/material-ui/guides/minimizing-bundle-size/) explain this unintuitive behavior

arthurvmdantas:
Oh oh, absolutely; I got it now. I should've looked into it before. 😅 😅 

Thanks for the clarification.

PR #657, opened 2024-05-23
use sha256 for short urls

tim-smart:


src/components/editor/rx/share.ts
```diff
@@ -64,22 +60,23 @@ const makeDefaultWorkspace = () =>
...
       const hash = get(hashRx)
       if (hash._tag === "None") return makeDefaultWorkspace()
       const compressed = yield* Effect.promise(() =>
         retrieveCompressed(hash.value)
       )
       if (!compressed) return makeDefaultWorkspace()
       const compression = yield* WorkspaceCompression
-      return yield* compression
-        .decompress({
+      return yield* _(
```
IMax153, line 71: Not that it really matters, but we can probably start removing the adapter usage when we see it.

tim-smart: I added it on purpose lol, sometimes the .pipe() method formats weird. Can replace with normal pipe.

IMax153: I figured it was probably on purpose lol

PR #760, opened 2024-07-10
add effect 3.5 release post

tim-smart:


content/blog/0035-effect-3.5.mdx
```diff
@@ -0,0 +1,124 @@
...
+for Cloudflare's D1 database.
+
+## Added RcRef & RcMap
+
+`RcRef` and `RcMap` are new reference counted types that can be used to manage resources.
+
+The wrapped resource will be acquired on the first access and released when no longer in use.
+
+`RcRef` can be used to manage a single resource, and `RcMap` can be used to manage
+multiple resources reference by a key.
```
fubhy, line 41:
```suggestion
multiple resources referenced by a key.
```

content/blog/0035-effect-3.5.mdx
```diff
@@ -0,0 +1,124 @@
...
+  yield* RcMap.get(map, "foo").pipe(
+    Effect.andThen(RcMap.get(map, "foo")),
+    Effect.scoped
+  )
+})
+```
+
+## Added Logger.pretty
+
+`Logger.pretty` is a new logger that leverages the features of the `console` APIs to provide a more visually appealing output.
```
fubhy, line 65: Let's add a note that this will be the default logger in Effect 4.0+

content/blog/0035-effect-3.5.mdx
```diff
@@ -0,0 +1,124 @@
...
+class MyError extends Data.Error<{ cause: Error }> {}
+```
+
+If you `Effect.log` a `Cause` containing an error with a `cause` property, it will now be
+visible in the log output.
+
+## @effect/sql-d1
+
+The `@effect/sql-d1` package has been released. This package provides `@effect/sql` support
+for Cloudflare's D1 database.
```
fubhy, line 32: Let's add a shoutout to ecyrbe here.

fubhy (changes requested):

content/blog/0035-effect-3.5.mdx
fubhy: Fair. Although I think there's a difference between changelog & commit attribution (which we already do) and acknowleding significant contributions by singling out (especially new) contributors for such in the release posts. I'd argue that by attributing everything in the release post we would actually water down the significance of such a shoutout. I think it'd be a nice gesture.

fubhy: Thoughts @schickling @mikearnaldi ?

mikearnaldi: Would be a first, I don't see the need imho

fubhy: Fine. Let's take the broader discussion around that elsewhere for now.

PR #762, opened 2024-07-10
Fix shorten service types

lokhmakov:
KeyValueStore was added to shorten, which caused a PlatformError. Without type guard, catchIf returns an invalid type, which causes a type error in Router rpcs.ts.

### DIFF
--- a/src/services/Shorten.ts
+++ b/src/services/Shorten.ts
@@ -11,7 +11,7 @@ const make = Effect.gen(function* (_) {
   const store = KVS.prefix(kvs, "shorten/")
 
   const shorten = (thing: string) =>
-    Effect.gen(function* (_) {
+    Effect.gen(function* () {
       if (thing.length > constMaxSize) {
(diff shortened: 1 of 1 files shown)


IMax153:
@lokhmakov - thanks for the PR! But I apologize, I'm not seeing the type error you are describing.

The type of `Shorten.shorten` is correct:

```ts
Effect<string, ShortenError, Shorten>
```

lokhmakov:
@IMax153 I see this:

image

And this:

image

IMax153:
@lokhmakov - perhaps something about your local environment is interfering? Have you tried restarting the TS server and / or reloading VSCode?

image

image

lokhmakov:
@IMax153 Yep, TS server restarted. I saw this error like couple of weeks ago. But right now I tried to setup in fresh gitpod and there is no such error. I will try to find what happens. Thanks.

IMax153:
@lokhmakov - no problem 👍

PR #797, opened 2024-08-02
fix (flake): use same node version for corepack

sectore:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

- `flake.nix`: 
   - Use `corepack` of `nixpkgs
   - Remove `flake-utils`
   - Remove uneeded `bun` + `nodePackages.json` from `buildInputs`
- `flake.lock`: Update by running `nix flake update`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,9 +17,10 @@
   }:
     flake-utils.lib.eachDefaultSystem (system: let
       pkgs = nixpkgs.legacyPackages.${system};
+      node = pkgs.nodejs_20;
       corepackEnable = pkgs.runCommand "corepack-enable" {} ''
         mkdir -p $out/bin
-        ${pkgs.nodejs-18_x}/bin/corepack enable --install-directory $out/bin
+        ${node}/bin/corepack enable --install-directory $out/bin
       '';
     in {
       formatter = pkgs.alejandra;
@@ -28,7 +29,7 @@
         default = pkgs.mkShell {
           buildInputs = with pkgs; [
             bun
-            nodejs_20
+            node
             corepackEnable
             nodePackages.json
           ];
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -17,9 +17,10 @@
   }:
     flake-utils.lib.eachDefaultSystem (system: let
       pkgs = nixpkgs.legacyPackages.${system};
+      node = pkgs.nodejs_20;
       corepackEnable = pkgs.runCommand "corepack-enable" {} ''
```
IMax153, line 21: As it turns out the `corepack` package in `nixpkgs` already enables itself if included in the derivation, so if you're up for it @sectore you can just copy what we did in the `flake.nix` in the [`effect` repository](https://github.com/Effect-TS/effect/blob/main/flake.nix).

IMax153 (changes requested):

flake.nix
sectore: Done 382024b . 

Not sure if `bun` is still needed here https://github.com/Effect-TS/website/blob/382024b31e0b1adea198fd8e1833b43c65d2bc7c/flake.nix#L18 There is a [pnpm-lock.yaml](https://github.com/Effect-TS/website/blob/main/pnpm-lock.yaml) in the repo only.  (just let me know)

IMax153: I believe both bun and nodePackages.json can be removed.

sectore: Done 9e7068d

IMax153:
@sectore - apologies that this got lost. Thank you for the PR 🙏 - will merge once CI builds.

IMax153 (approved):

PR #821, opened 2024-08-26
Add section "The flow method"

thanhnguyen2187:
I think the method might be useful but under-documented and hence the PR

### DIFF
--- a/content/docs/400-guides/100-essentials/600-pipeline.mdx
+++ b/content/docs/400-guides/100-essentials/600-pipeline.mdx
@@ -531,6 +531,34 @@ const program = Effect.all([fetchTransactionAmount, fetchDiscountRate]).pipe(
 )
 ```
 
+## The flow method
+
+Sometimes, you may need to create a reusable pipeline of operations. In such
+cases, you can use the `flow` method.
+
+```ts twoslash
+import { Effect, Duration, flow } from "effect"
+
+const myRetryLogic = flow(
+  Effect.timeout(Duration.seconds(3)),
+  Effect.tapError((error) => Effect.log(`Failed: ${error}`)),
+  Effect.retry({ times: 2 }),
+  Effect.tapError((error) => Effect.log(`Abandon as retry failed: ${error}`)),
+)
+
+const fetchTransactionAmount = Effect.promise(() => Promise.resolve(100))
+const fetchDiscountRate = Effect.promise(() => Promise.resolve(5))
+
+const fetchTransactionAmountWithRetry = fetchTransactionAmount.pipe(myRetryLogic)
+const fetchDiscountRateWithRetry = fetchDiscountRate.pipe(myRetryLogic)
+```
+
+<Info>
+    `flow` and `pipe` are very similar. The main difference is that `flow` returns *a function that returns an Effect*
+    (`() => Effect`), while `pipe` returns an `Effect`. It suggests that `pipe` should be used when you want a result,
+    while `flow` should be used for further composition.
+</Info>
+
 ## Cheatsheet
 
 Let's summarize the transformation functions we have seen so far:


content/docs/400-guides/100-essentials/600-pipeline.mdx
```diff
@@ -531,6 +531,34 @@ const program = Effect.all([fetchTransactionAmount, fetchDiscountRate]).pipe(
...
+
+const fetchTransactionAmount = Effect.promise(() => Promise.resolve(100))
+const fetchDiscountRate = Effect.promise(() => Promise.resolve(5))
+
+const fetchTransactionAmountWithRetry = fetchTransactionAmount.pipe(myRetryLogic)
+const fetchDiscountRateWithRetry = fetchDiscountRate.pipe(myRetryLogic)
+```
+
+<Info>
+    `flow` and `pipe` are very similar. The main difference is that `flow` returns *a function that returns an Effect*
```
nspaeth, line 557: This isn't quite right. `flow` doesn't always return an `Effect`, it returns whatever the return type of the last function in the pipeline is.

content/docs/400-guides/100-essentials/600-pipeline.mdx
```diff
@@ -531,6 +531,34 @@ const program = Effect.all([fetchTransactionAmount, fetchDiscountRate]).pipe(
...
+    (`() => Effect`), while `pipe` returns an `Effect`. It suggests that `pipe` should be used when you want a result,
```
nspaeth, line 558: I think I simplified example of how the same operations can be performed with `pipe` and `flow`, would be help highlight that `flow` is just a shorthand for `(...args) => pipe(yourInitialOp(...args), ...otherOps)`

nspaeth: Maybe something like:
```typescript
import { pipe, flow, String, Array } from "effect"
type Name = { first: string, last: string }
// const greetFirstNamesFlow: (names: Name[], numberOfPeopleToGreet: number) => string
const greetFirstNamesFlow = flow(
  (names: Name[], numberOfPeopleToGreet: number) => Array.take(names, numberOfPeopleToGreet),
  Array.map(name => name.first),
  Array.map(first => `Hello + ${first}`),
  Array.join('\n'),
)

// const greetFirstNamesPipe: (names: Name[], numberOfPeopleToGreet: number) => string
const greetFirstNamesPipe = (names: Name[], numberOfPeopleToGreet: number) =>
  pipe(
    Array.take(names, numberOfPeopleToGreet),
    Array.map(name => name.first),
    Array.map(first => `Hello + ${first}`),
    Array.join('\n'),
  )
```

content/docs/400-guides/100-essentials/600-pipeline.mdx
thanhnguyen2187: Yep thanks for the clarification. I'll try to correct it

content/docs/400-guides/100-essentials/600-pipeline.mdx
```diff
@@ -60,6 +60,27 @@ In the above example, we start with an input value of `5`. The `increment` funct
 
 The result is equivalent to `subtractTen(double(increment(5)))`, but using `pipe` makes the code more readable because the operations are sequenced from left to right, rather than nesting them inside out.
 
+## flow
+
+`flow` is a "lazier" version of `pipe`. While `pipe` returns a concrete value, `flow` returns a function that returns a
+result. It allows us to reuse the same composed functions in different contexts.
```
thanhnguyen2187, line 66: @nspaeth I reread the whole document again, and found that there is duplication `pipe` explanation. Therefore I added the `flow` function along with the `pipe` function, and redid the end, where everything is tied together a bit

image

PR #898, opened 2024-11-05
Use encoding module

fubhy:


IMax153 (approved):
Just tested this on the preview site and it all looks like it's working well!

Thanks @fubhy 👍

PR #1046, opened 2025-01-31
Migrating an endpoint that uses a database.

samwho:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

**Currently still a work in progress.**


content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,262 @@
...
+// => "Hello, world!"
+```
+
+What if you want to run an Effect that calls another Effect? You can use
+`Effect.gen` to do that:
+
+```ts twoslash {2-5}
+import { Effect } from "effect"
+// ---cut---
+const promise = Effect.promise(async () => "Hello, world!")
```
IMax153, line 132: Might be worth renaming this from `promise` to something else to avoid a user interpreting the variable as a `Promise` instead of an `Effect`.

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,195 @@
...
+)
+
+NodeRuntime.runMain(
+  Layer.launch(
+    Layer.provide(
+      HttpServer.serve(router),
+      NodeHttpServer.layer(() => createServer(app), { port: 3000 }),
+    ),
+  ),
+)
```
IMax153, line 53: Since you've already taught folks about `.pipe`, you could re-write this to make it flow sequentially:

```ts
HttpServer.server(router).pipe(
  Layer.provide(NodeHttpServer.layer(() => createServer(app), { port: 3000 })),
  Layer.launch,
  NodeRuntime.runMain
)
```

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,195 @@
...
+sqlite> INSERT INTO users (name) VALUES ("Bob");
+```
+
+Then you can run the server with:
+
+```shell
+$ bun index.ts
+```
+
+And in other shell we can test the endpoint:
```
IMax153, line 71:
```suggestion
And in another shell we can test the endpoint:
```

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,195 @@
...
+  HttpRouter.get("/health", health),
+  HttpRouter.get("/users", users),
+)
+```
+
+This no longer gives us a type error. We've split our computation into two
+parts: fetching the data with `Effect.sync`, and serializing it with
+`HttpServerResponse.json`. The call to `pipe` takes the result of the first
+Effect and passes it to `HttpServerResponse.json`, flattening any nesting in
+the process.
```
IMax153, line 195: This is not correct - `body` in this case has the type `Effect<{ users: Array<string> }>`, which we're now passing to `HttpServerResponse.json`. Because `HttpServerResponse.json` takes data of type `unknown`, you're going to attempt to serialize an `Effect` instead of the `body`. 

The reason you don't get a type error here is because the input to `HttpServerResponse.json` is `unknown`, which will allow any value.

At this point, you may choose to introduce the concept of sequencing `Effect`s.

```ts
const users = Effect.sync(() => {
  return { users: db.prepare("SELECT * FROM users").all() }
}).pipe(
  // Similar to `Effect.map`, `Effect.flatMap` receives the value of the previous effect 
  // but instead of returning a new _value_, you return a new _Effect_
  Effect.flatMap((body) => HttpServerResponse.json(body))
  // Or alternatively `Effect.andThen((body) => ...)` if you prefer that naming
)
// `users` is type `Effect<HttpServerResponse, HttpBodyError>`
```

or you can introduce `Effect.suspend`, which is the same as `Effect.sync`, except instead of taking a side effect that returns a value, it takes a side effect that returns an `Effect`:

```ts
const users = Effect.suspend(() => {
  // Side effect
  const body = { users: db.prepare("SELECT * FROM users").all() }
  // return an effect
  return HttpServerResponse.json(body)
})
// `users` is type `Effect<HttpServerResponse, HttpBodyError>`
```

IMax153 (changes requested):

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
samwho: With the code as-is, e.g.

```typescript
const users = Effect.sync(() => {
  return {
    users: db.prepare("SELECT * FROM users").all(),
  };
}).pipe((body) => HttpServerResponse.json(body))
```

If I make a request to `localhost:3000/users` I get the following:

```
❯ http get "localhost:3000/users"
HTTP/1.1 200 OK
Content-Length: 57
Content-Type: application/json; charset=utf-8
Date: Fri, 07 Feb 2025 10:15:32 GMT
ETag: W/"39-6Qu85qIU12mhgGYWwHErQVfJRxI"
X-Powered-By: Express

{
    "users": [
        {
            "id": 1,
            "name": "Alice"
        },
        {
            "id": 2,
            "name": "Bob"
        }
    ]
}
```

So I made the assumption it was doing something clever with the `Effect` it receives. 

The result of using flatMap, e.g.

```typescript
const users = Effect.sync(() => {
  return {
    users: db.prepare("SELECT * FROM users").all(),
  };
}).pipe(Effect.flatMap(HttpServerResponse.json));
```

Is identical.

```
❯ http get "localhost:3000/users"
HTTP/1.1 200 OK
Content-Length: 57
Content-Type: application/json; charset=utf-8
Date: Fri, 07 Feb 2025 10:18:23 GMT
ETag: W/"39-6Qu85qIU12mhgGYWwHErQVfJRxI"
X-Powered-By: Express

{
    "users": [
        {
            "id": 1,
            "name": "Alice"
        },
        {
            "id": 2,
            "name": "Bob"
        }
    ]
}
```

mikearnaldi: Looks like a bug in the impl of `json` probably something uses andThen and it ends up flattening when it shouldn't, cc @tim-smart

samwho: Sorry, disregard what I'm saying here. I feel silly. I had forgotten to delete the Express handler for the route, and that's what was running. Not the Effect code. There's no bug to fix here, it was failing as expected.

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,433 @@
...
+
+This prints the following, with `result` still being 8:
+
+```
+1 + 2 = 3
+  * 3 = 9
+  - 1 = 8
+```
+
+As we migrate more complex endpoints, we'll see how this ability to "tap into"
```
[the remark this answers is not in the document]
samwho: Thank you 🙏 

I'm trying to use language that might be familiar to people coming from other languages, e.g. Ruby has a `tap` method that works similarly: https://ruby-doc.org/3.4.1/Kernel.html#method-i-tap.

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,395 @@
...
+it returns when it fails, and we'll talk about `Requirements` a bit later.
+
+When we created our example Effects in the previous section, these all took on
+the type `Effect<string, never, never>`.
+
+```ts twoslash
+import { Effect } from "effect"
+//---cut---
+Effect.sync(() => "Hello, world!")
+//=> Effect.Effect<string, never, never>
```
IMax153, line 183: Just as an FYI, if you want to use Twoslash to display the type of something you can use the `// ^?` query operator:

```suggestion
import { Effect } from "effect"
//---cut---
const _ = Effect.sync(() => "Hello, world!")
//        ^?
```

I see you're using it in the snippet below, so I assume this was deliberate 😄

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,395 @@
...
+This is what the type error from earlier was trying to tell us. Here's the
+error again:
+
+```text wrap
+Argument of type 'Effect<Effect<HttpServerResponse, HttpBodyError, never>, never, never>' is not assignable to parameter of type 'Handler<never, never>'
+```
+
+Under the hood, a `Handler<never, never>` is an `Effect<HttpServerResponse,
+never, never>`. Because `Effect<HttpServerResponse, HttpBodyError, never>` is
+not an `HttpServerResponse`, the type checker gets upset with us.
```
IMax153, line 215:
```suggestion
Under the hood, a `Handler<never, never>` is just a type alias for `Effect<HttpServerResponse,
never, never>`. Because `Effect<HttpServerResponse, HttpBodyError, never>` is
not an `HttpServerResponse`, the type checker gets upset with us.
```

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,395 @@
...
+  Effect.map(({ rows }) => ({ users: rows })),
+  Effect.flatMap(HttpServerResponse.json)
+);
+```
+
+This no longer gives us a type error. We've split our computation into parts:
+fetching the data with `Effect.promise`, manipulating its shape with
+`Effect.map`, and serializing it with `HttpServerResponse.json`. The
+`Effect.flatMap` at the end care of the accidental `Effect` nesting.
+
```
IMax153, line 258:
```suggestion
`Effect.map`, and serializing it with `HttpServerResponse.json`. The
`Effect.flatMap` at the end takes care of the accidental `Effect` nesting we had in our previous implementation.

```

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,395 @@
...
+As we migrate more complex endpoints, we'll see how this ability to "tap into"
+the pipeline is really useful.
+
+## Adding error handling to our database query
+
+All this talk of error handling might have gotten you thinking: can't our
+database query fail? It can! If the query is malformed, `db.query` will throw an
+error. When that happens, at the moment our endpoint will return a 500 error to
+the client. Change the SQL query to `SELECT * FROM nonexistnet` and try it:
```
IMax153, line 329:
```suggestion
All this talk of error handling might have gotten you thinking: can't our
database query fail? It can! If the query is malformed, `db.query` will throw an
error. When that happens, our endpoint will currently return a 500 error to
the client. Change the SQL query to `SELECT * FROM nonexistent` and try it:
```

IMax153 (commented):
Just a few minor comments for today! Great work @samwho - it's coming along very nicely!

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,515 @@
...
+import { Effect } from "effect"
+
+const effect = Effect.succeed(1).pipe(
+  Effect.tryMap({
+    try: (n) => n + 1,
+    catch: () => new Error(),
+  }),
+)
+
+const noError: Effect.Effect<number, never, never> = effect
```
IMax153, line 426: This transition felt a bit abrupt to me. To me, it might feel more natural if you insert one of the "aside"-esque comments you've had previously (recall the `pipe` aside). Perhaps "An aside on errors in Effect" or something?

Also, before now `Effect.tryMap` has not even been mentioned but now we're using it all of a sudden. 

Might be worth using the user fetch code from above for these samples instead of the random `tryMap` code, given that you also insert it into the final snippet later anyways.

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,515 @@
...
+  }),
+  Effect.orDie
+)
+
+const noError: Effect.Effect<number, Error, never> = effect
+```
+
+This effectively recreates the behaviour of throwing an exception. Defects are
+critical errors that you don't want to handle. If you were to create a defect in
+an Effect handler, the result would be a 500 error returned to the user.
```
IMax153, line 467: Might be worth one more sentence or something here providing a clear distinction between "expected" errors, which are part of the program's domain, and "defects" which are critical errors (as you explained).

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,515 @@
...
+
+```ts twoslash {1,4-5}
+import { HttpServerResponse } from "@effect/platform"
+import { Effect, Layer } from "effect"
+import { createServer } from "node:http"
+import { Client} from "pg"
+const db = new Client({ user: "postgres" })
+//---cut---
+class UserFetchError extends Error {}
+
```
IMax153, line 389: For a variety of reasons, I would advise against this pattern. 

For one, the lack of a discriminator in the type (e.g. a `_tag`) means that another error which extends `Error` can be directly assignable to the `UserFetchError`: https://effect.website/play#55cd56ec464d.

My recommendation would be to either:

1. Take the time to introduce creating your own domain-specific errors more properly here via `Data.TaggedError` https://effect.website/docs/data-types/data/#taggederror
2. Just make the error an interface in this case for now if you want to keep it simple

```ts
interface UserFetchError {
  readonly _tag: "UserFetchError"
  readonly message: string
  readonly cause: unknown
}
```

I personally think it might be good to go for the first option given that `Data.TaggedError` will automatically attach a `_tag` to our error type and also extends `Error` for us by default, but I leave that decision to you.

IMax153 (changes requested):

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
samwho: I agonised over this, because tags do feel like they deserve special attention and I'll soon get to them when I talk about services. I wanted to wait for a place I _needed_ to use them in error handling to introduce them. I also don't want to wander too far away from the purpose of this section: to migrate an endpoint that calls an external API.

This section in its entirety is still very early, probably the earliest I've pushed a section up for review. I'm not tightly committed to any part of it. If you think we really should introduce tags here, I don't think it would be impossible to do so. I do think we'll have to cut something else to make room for them, though, and it might have to be services.

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
samwho: I did wonder about whether to explicitly call out `Effect.tryMap` and landed on not doing so. It felt like because we've already covered `Effect.map`, and we looked at `Effect.try` in the previous section, the reader wouldn't flinch at `Effect.tryMap` being casually used without explanation.

A small nod to it might be the way to go, though. I need to start being careful, I've not spent more time with Effect than the reader and am likely prone to thinking they'll understand things they might not. 😅

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
IMax153: Totally understand, and honestly I don't want to impose my opinion too much on the direction you take here.

My only advice would be that if you want to avoid introducing `Data.TaggedError`, etc., maybe just add a `_tag` property or something to your custom error type and do another one of those "this is important but we'll discuss it later" things. Just to ensure you have a discriminator for the type in place, and you don't create an unintentional footgun with multiple error types that are assignable to one another.

content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,355 @@
...
+The Effect equivalent of our Express `/health` handler looks like this:
+
+```ts twoslash showLineNumbers=false
+import { Effect } from "effect"
+// ---cut---
+import { HttpServerResponse } from "@effect/platform"
+
+const health = Effect.sync(() => {
+  return HttpServerResponse.text("ok")
+})
```
IMax153, line 89:
```suggestion
const health = HttpServerResponse.text("ok")
```

I still think we should get rid of the `Effect.sync` wrapper here. It's superfluous since the code works without it, and adds one more thing you have to explain.

content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,355 @@
...
+)
+console.log(syncResult)
+// => "Hello, world!"
+
+const asyncResult = await Effect.runPromise(
+  Effect.promise(async () => "Hello, world!")
+)
+console.log(asyncResult)
+// => "Hello, world!"
+```
```
IMax153, line 159: Despite the commentary we received, I would advocate strongly **against** showing `Effect.runSync` at all. A user shouldn't care at all about function coloring. Any `Effect` can be run with `Effect.runPromise`, and we should probably just stick to this during the tutorial.

content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,355 @@
...
+const result = await Effect.runPromise(gen)
+console.log(result)
+// => "HELLO, WORLD!"
+```
+
+This makes use of JavaScript's [generator functions][1]. The way Effect uses
+them, if you think of an Effect as being like a `Promise`, `yield*`ing an
+`Effect` is the same as `await`ing a `Promise`.  Under the hood, all `Effect`s
+are generators, and `yield*`ing one passes it to the Effect runtime and waits
+for the result.
```
IMax153, line 181:
```suggestion
This makes use of JavaScript's [generator functions][1]. If you think of an Effect as similar to a `Promise`, `yield*`ing an `Effect` is the same as `await`ing a `Promise`.  Under the hood, all `Effect`s are generators, and `yield*`ing one passes it to the Effect runtime and waits for the result.
```

content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,355 @@
...
+it a lot. Another thing you'll see used a lot is `Effect.pipe`. Here's the same
+code as above, but using `Effect.pipe`:
+
+```ts twoslash showLineNumbers=false
+import { Effect } from "effect"
+// ---cut---
+const effect = Effect.sync(() => "Hello, world!").pipe(
+  Effect.map((s) => s.toUpperCase())
+)
+const result = await Effect.runSync(effect)
```
IMax153, line 193:
```suggestion
const result = await Effect.runPromise(effect)
```

content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,355 @@
...
+})
+// ---cut---
+const router = HttpRouter.empty.pipe(
+  HttpRouter.get("/health", health),
+)
+```
+
+`HttpRouter` is a data structure that represents a collection of routes. The
+simplest router is one with no routes at all, and Effect exposes that to us as
+the `HttpRouter.empty` value. Under the hood, this is itself an `Effect`:
```
IMax153, line 241:
```suggestion
the `HttpRouter.empty` value. Under the hood, an `HttpRouter` is itself an `Effect`:
```

content/src/content/docs/learn/tutorials/first-endpoint.mdx
```diff
@@ -0,0 +1,355 @@
...
+)
+
+// vs
+
+HttpRouter.empty.pipe(
+  HttpRouter.get("/version", version),
+  HttpRouter.get("/status", status),
+  HttpRouter.get("/health", health),
+)
+```
```
IMax153, line 297: Suggest two snippets here. The first one could leads with "... see how this gets out of hand..." and the second leads with something like "versus using `pipe`":

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,466 @@
...
+```shell showLineNumbers=false
+docker run -d -e POSTGRES_HOST_AUTH_METHOD=trust -p 5432:5432 postgres:17
+```
+
+<Aside type="note" title="I can't use Docker!">
+If you aren't able to run Docker locally, for example because your work's laptop
+policy doesn't allow it, an alternative is to use a hosted Postgres service.
+For example, [Neon](https://neondb.com/) offers a free tier that should be
+more than enough for this tutorial. They also offer a web-based UI for creating
+tables and rows.
```
IMax153, line 29: While I understand wanting to point users in the right direction, I'm not sure we should direct them to a _specific_ hosted service.

content/src/content/docs/learn/tutorials/hooking-up-a-database.mdx
```diff
@@ -0,0 +1,466 @@
...
+const health = Effect.sync(() => {
+  return HttpServerResponse.text("ok")
+})
+
+const router = HttpRouter.empty.pipe(
+  HttpRouter.get("/health", health)
+)
+
+const main = Effect.gen(function* () {
+  yield* Effect.promise(() => db.connect())
```
IMax153, line 93:
```suggestion
  // Make sure we connect to the database before starting the server
  yield* Effect.promise(() => db.connect())
```

Might be worth adding a comment just for this  snippet since user's might be a bit confused why this is required? Idk

content/src/content/docs/learn/tutorials/calling-an-api.mdx
```diff
@@ -0,0 +1,400 @@
...
+const query = Effect.gen(function* () {
+  //  ^?
+  const req = yield* HttpServerRequest.HttpServerRequest
+  return yield* HttpServerResponse.json({ url: req.url })
+})
+```
+
+An Effect that returns an `HttpServerResponse`, may throw an `HttpBodyError`,
+but also, for the first time, has something other than `never` in that third
+spot. As a reminder, this is the `Requirement` type. What it's saying here is
```
IMax153, line 147:
```suggestion
spot. As a reminder, this is the `Requirements` type. What it's saying here is
```

IMax153 (commented):
Leaving my feedback on `Providing Requirements` onward off this review for now, as discussed in Discord.

content/src/content/docs/learn/tutorials/calling-an-api.mdx
```diff
@@ -0,0 +1,452 @@
...
+const app = express()
+
+app.get("/query", async (req, res) => {
+  const query = req.query.q
+  if (typeof query !== "string") {
+    res.status(400).send("Query parameter 'q' is required")
+    return
+  }
+
+  const url = `https://en.wikipedia.org/w/rest.php/v1/search/page?q=${query}&limit=1`
```
IMax153, line 32:
```suggestion
  const baseUrl = "https://en.wikipedia.org/w/rest.php/v1/search/page`
  const queryParams = `?q=${query}&limit=1`
  const url = `${baseUrl}${queryParams}`
```
To avoid horizontal scroll when the screen is wide enough to avoid it.

content/src/content/docs/learn/tutorials/calling-an-api.mdx
```diff
@@ -0,0 +1,452 @@
...
+```
+
+To:
+
+``` showLineNumbers=false
+Effect<void, never, never>
+```
+
+So each Effect can be thought of as a self-contained program. Its requirements
+are its inputs, its success value is its output, and it can error. Effect forces
```
IMax153, line 273: Technically speaking, the output of an Effect program can be either its success value or its possible errors.

content/src/content/docs/learn/tutorials/services-and-config.mdx
```diff
@@ -0,0 +1,381 @@
...
+    getUsers: () => Effect.promise(async () => {
+        const { rows } = await db.query("SELECT * FROM users")
+        return rows
+    }),
+})
+```
+
+The choice of `DatabaseLive` for the name is a common convention you will see
+in Effect code, where "live" is referring to the fact that this is the version
+of this service we will use in live, production code.
```
IMax153, line 77: We've sort of been re-thinking the whole "Live" nomenclature because it's not intuitive for JS devs. We've been starting to advocate more for `<ServiceName>Layer` as a convention, so it's immediately clear to users what the variable is (in this case a `Layer`).

content/src/content/docs/learn/tutorials/services-and-config.mdx
```diff
@@ -0,0 +1,381 @@
...
+`Effect.tryPromise` and wrapping any error we get back from `pg`.
+
+```ts twoslash ins={1-3,8,15-19} showLineNumbers=false
+import { Context, Effect } from "effect";
+import { Client } from "pg"
+const db = new Client({ user: "postgres" })
+//---cut---
+class DatabaseError extends Error {
+  readonly _tag = "DatabaseError";
+}
```
IMax153, line 90: It would be worth adding a `constructor` here to take the `cause`:
```ts
constructor(readonly cause: unknown) {}
```

samwho: JavaScript's default `Error` class already accepts a cause as part of its second argument:

```typescript
class DatabaseError extends Error {
  readonly _tag = "DatabaseError";
}

new DatabaseError("oh no", { cause: new Error("cause") });
```

IMax153: Wow! TIL! 👍

IMax153: Wow! TIL! 👍

PR #1076, opened 2025-03-05
podcast/episode_#2

DadeSko:
add episode 2


DadeSko:
Hey @IMax153, here's the draft for the podcast page! 

I also managed to run the srt to json script, so everything should be in place. 
The metadata are not updated yet 'cuz we'll need the rss feed from transistor after the episode is released. 

Let me know if there's any issue and thanks a lot for the help!

PR #1123, opened 2025-04-24
add autosave

titouancreach:
## Type

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

add auto save to the playground

## Related

- Related Issue #
- Closes #

### DIFF
--- a/content/src/components/editor/components/reset-button.tsx
+++ b/content/src/components/editor/components/reset-button.tsx
@@ -1,0 +1,18 @@
+import { useRxSet } from "@effect-rx/rx-react"
+import { Button } from "@/components/ui/button"
+import { useWorkspaceHandle } from "../context/workspace"
+
+export function ResetButton() {
+  const handle = useWorkspaceHandle()
+  const reset = useRxSet(handle.reset)
+
+  return (
+    <Button
+      variant="ghost"
+      className="z-10 bg-[--sl-color-bg-nav] hover:bg-[--sl-color-gray-6] dark:hover:bg-[--sl-color-gray-5] border border-[--sl-color-text] text-[--sl-color-white] cursor-pointer flex-none"
+      onClick={() => reset()}
+    >
+      Reset
+    </Button>
+  )
+}
--- a/content/src/components/editor/rx/editor.ts
+++ b/content/src/components/editor/rx/editor.ts
@@ -13,17 +13,21 @@ import { Loader } from "../services/loader"
 import { Monaco } from "../services/monaco"
 import { WebContainer } from "../services/webcontainer"
 import type { RxWorkspaceHandle } from "./workspace"
+import { AutoSave } from "../services/auto-save"
 
 export const editorThemeRx = Rx.map(themeRx, (theme) =>
   theme === "dark" ? "dracula" : "vs"
 )
 
-const runtime = Rx.runtime(Layer.mergeAll(
-  Loader.Default,
-  Monaco.Default,
-  Toaster.Default,
-  WebContainer.Default
-)).pipe(Rx.setIdleTTL("10 seconds"))
+const runtime = Rx.runtime(
+  Layer.mergeAll(
+    Loader.Default,
+    Monaco.Default,
+    Toaster.Default,
+    WebContainer.Default,
+    AutoSave.Default
+  )
+).pipe(Rx.setIdleTTL("10 seconds"))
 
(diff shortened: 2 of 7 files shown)


content/src/components/editor/rx/editor.ts
```diff
@@ -60,43 +65,76 @@ export const editorRx = Rx.family((handle: RxWorkspaceHandle) => {
...
+          ),
+          Stream.ensuring(
+            Effect.suspend(() => {
+              const content = editor.editor.getValue()
+              if (content.trim().length === 0) {
+                return Effect.void
+              }
+              return container.writeFile(fullPath, content, file.language)
+            })
+          )
```
IMax153, line 114: @titouancreach - there are a lot of these unnecessary formatting changes in this PR, which makes it difficult to parse what the actual changes to the codebase are.

Could you please remove an ancillary formatting changes?

content/src/components/editor/components/reset-button.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function ResetButton() {
+  const handle = useWorkspaceHandle()
+  const reset = useRxSet(handle.reset)
+
+  return (
+    <Button
+      variant="ghost"
+      className="z-10 bg-[--sl-color-bg-nav] hover:bg-[--sl-color-gray-6] dark:hover:bg-[--sl-color-gray-5] border border-[--sl-color-text] text-[--sl-color-white] cursor-pointer flex-none"
+      onClick={() => reset()}
+      title="Réinitialiser l'éditeur"
```
IMax153, line 14: The editor only supports English for now.

While we don't currently support internationalization, it's certainly something we'd like to do in the future.

IMax153 (changes requested):

content/src/components/editor/components/reset-button.tsx
titouancreach: ohhhh cursor added this I didn't even notice :D

content/src/components/editor/rx/editor.ts
titouancreach: Yes sure ! I used format and save with prettier since I have seen a prettierrc file !

titouancreach:
I hope I didn't mess up too much with git but I think I removed all the unwanted diffs !

content/src/components/editor/rx/workspace.ts
```diff
@@ -122,6 +127,13 @@ export const workspaceHandleRx = Rx.family((workspace: Workspace) =>
         )
       )
 
+      const reset = Rx.fn(() =>
+        Effect.gen(function* () {
+          yield* autosave.resetLocal("save")
+          window.location.reload()
```
IMax153, line 133: @titouancreach - this looks really good! The only question I have is whether the reload here is necessary? Just want to make sure I understand the logic before I approve / merge the PR.

titouancreach: tbh there is probably another way (better way) to reset the workspace without reloading the page. Reloading the page is probably more "pragmatic" (less side effects maybe). Do you want me to find a way to avoid page reload ?

This code juste remove the localstorage item and reload the page.
Maybe, we can add a timestamp or something because if you reopen the playground 2 weeks after, you probably don't want it to start with your old code

titouancreach: Got it

titouancreach:
Hello @IMax153,
I experimented some things to hot swap current workspace: 
  - directly update the SubscriptionRef that contains the workspace with `SubscriptionRef.update`
  - update the Workspace value used in the WorkspaceProvider 
  
 both method failed with weird error like (file alread exist blabla).
 
 Can you point me out the right direction ?

IMax153:
> both method failed with weird error like (file alread exist blabla).

This means that you're trying to write a file in the web container that is already there. Probably the fix here is to recursively delete all files under the workspace folder and then do either of the methods you proposed above.

IMax153:
@titouancreach - I apologize but I believe @tim-smart has gone ahead and implemented this functionality in #1131 to kick the tires on a new version of `effect-rx`. I do believe that some of this code was inspiration though.

If you're curious, it looks like he elected to use the `BrowserKeyValueStore` to store the workspace files.

tim-smart:
Ah sorry I swear I left a comment, but it didn't show up.

I cleaned a few things up which broke this PR, so took a different strategy to implement auto save.

PR #1125, opened 2025-04-24
Fix managed runtime example type hovering

yatoogamii:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description

I think there is an artifact within the example of [managed runtime Effect.Tag](https://effect.website/docs/runtime/#effecttag) where the type of the `action` variable is always shown as if we were hovering over it

I just move the type in a proper comment above the variable based on other examples within the docs

before change

Screenshot from 2025-04-25 04-13-12
after change

Screenshot from 2025-04-25 04-12-57

## Related


gcanti:
Thanks @yatoogamii

PR #1162, opened 2025-06-18
docs: add nested error paths section to `Schema.filter`

giuliobracci:
## Type

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [X] Documentation Update

## Description

Adds a new `Schema` documentation example. It demonstrates using Schema.filter to report errors on nested fields with interdependent validation rules (e.g., conditional discount based on base price).


giuliobracci:
I agree with your point @gcanti.

However, what do you think about adding a note to the specifying-error-paths section, highlighting this possibility?

I've encountered situations in practice where properties in the root object needed to report a nested error. Porting the subschema isn't straightforward in such cases, as it was generated dynamically by a function.

PR #1168, opened 2025-07-01
Fix incorrect default namespace delimiter in configuration.mdx

johtso:


gcanti (commented):
AFAIK they are correct

if I run (on Mac)

```bash
SERVER_HOST=foo SERVER_PORT=8080 TIMEOUT=500 npx tsx repro.ts
```

I can see the successful message

```
Application started: foo:8080
```

EDIT:

```ts
// repro.ts
import { Config, Effect } from "effect"

class ServiceConfig {
  constructor(
    readonly host: string,
    readonly port: number,
    readonly timeout: number
  ) {}
  get url() {
    return `${this.host}:${this.port}`
  }
}

const serverConfig = Config.all([Config.string("HOST"), Config.number("PORT")])

const serviceConfig = Config.map(
  Config.all([
    // Read 'HOST' and 'PORT' from 'SERVER' namespace
    Config.nested(serverConfig, "SERVER"),
    // Read 'TIMEOUT' from the root namespace
    Config.number("TIMEOUT")
  ]),
  ([[host, port], timeout]) => new ServiceConfig(host, port, timeout)
)

const program = Effect.gen(function* () {
  const hostPort = yield* serviceConfig
  console.log(`Application started: ${hostPort.url}`)
})

Effect.runPromise(program)
```

PR #1173, opened 2025-07-03
Add TS+ Post-Mortem blog post

mikearnaldi:
## Summary

Add comprehensive retrospective blog post on the TS+ (TypeScript Plus) experiment - our year-long attempt to fork TypeScript for enhanced Effect developer experience.

## What's included

- **Background**: What TS+ was and our ambitious design goals
- **Achievements**: Technical breakthroughs we accomplished including fluent APIs, runtime type derivation, extensible methods, custom operators, enhanced do syntax, global imports, and native pipe operator
- **Scale**: How far we went with the experiment - deep compiler work, complete Effect ecosystem rewrite, production users on alpha releases
- **What went wrong**: Fundamental architecture mismatches, tooling ecosystem fragmentation, performance penalties, and limited value proposition
- **Lessons learned**: Key insights about runtime behavior, targeted compiler patches, LSP plugins, and respecting build-tooling boundaries
- **Future considerations**: Thoughts on TSGo, JSX as a potential frontier, and Effect's frontend opportunities

## Context

This post provides valuable insights into the challenges and learnings from attempting to enhance TypeScript at the language level, and explains our current approach of improving developer experience through intelligent tooling while maintaining ecosystem compatibility.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/content/src/content/docs/blog/ts-plus-postmortem.mdx
+++ b/content/src/content/docs/blog/ts-plus-postmortem.mdx
@@ -60,25 +60,23 @@ Perhaps most significantly, we successfully implemented automatic generation of 
 interface Person {
   name: string;
   age: Option<number>;
-  friends: Chunk<Person>;
 }
 
-// Using Effect Schema - the Derive() calls would generate all runtime operations
+// TS+ approach - automatic derivation from type definition
+export const PersonEncoder: Encoder<Person> = Derive();
+
+// Alternative: Using Effect Schema (modern approach)
 const PersonSchema = Schema.Struct({
   name: Schema.String,
   age: Schema.Option(Schema.Number),
-  friends: Schema.Array(Schema.suspend(() => PersonSchema))
 });
 
 export const PersonEncoder = Schema.encodeSync(PersonSchema);
-export const PersonDecoder = Schema.decodeUnknownSync(PersonSchema);
-export const PersonGuard = Schema.is(PersonSchema);
 
 // Usage in practice
 const encoded = PersonEncoder({
   name: "Mike",
-  age: Option.some(30),
-  friends: []
+  age: Option.some(30)
 });
 ```
 


content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+)
+
+// TS+ fluent API
+const program = IO.succeed(0)
+  .map(n => n + 1)
+  .flatMap(r => 
+    IO.succeedWith(() => {
+      console.log(`result: ${r}`);
+    })
+  );
```
IMax153, line 51: It might be worthwhile to have a comment here that indicates that this code, when compiled, outputs code similar to the above. On first read, while I know this is the case, I guarantee another reader will not.

content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+interface Person {
+  name: string;
+  age: Option<number>;
+  friends: Chunk<Person>;
+}
+
+// Automatically generated codecs, decoders, guards
+export const PersonEncoder: Encoder<Person> = Derive();
+export const PersonDecoder: Decoder<Person> = Derive();
+export const PersonGuard: Guard<Person> = Derive();
```
IMax153, line 69: Again, I think it is worth "highlighting" the `Derive` call here somehow to the reader's attention to the fact that all the code will be generated. Perhaps even showing a small snippet of the code that might be generated (even if it's not accurate).

I think showing examples of the code generation will help the reader to understand how impactful this is.

content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+**Extensible Method Definitions**: Created a system where developers could add methods to existing types without modifying original classes:
+
+```typescript
+/**
+ * @tsplus type IO
+ */
+export class IO<A> {
+  constructor(readonly io: () => A) {}
+}
```
IMax153, line 87: I would suggest giving an example below the bullet so the reader knows what they are looking at.

Something like:
> The `@tsplus type IO` directive registers the `IO` type with TS+ under the name `IO`. Once that is done, other directives can be used to augment the `IO` type. For example, `@tsplus fluent IO map` registers the `map` function as a fluent method on the `IO` type, allowing APIs to be chained together without attaching them to the same runtime object.

I can't remember if the compiler gave us errors when we mistyped these directives, but if it did might be worth also mentioning that here.

content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * @tsplus fluent IO flatMap
+ */
+export function flatMap<A, B>(self: IO<A>, f: (a: A) => IO<B>): IO<B> {
+  return new IO(() => f(self.io()).io());
+}
+
+// Usage becomes natural - methods appear on the type:
```
IMax153, line 103:
```suggestion
// Usage is intuitive - fluent methods are discoverable via autocomplete
```

content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+// Usage: combining IOs with natural operator syntax
+const zipped = IO.succeed(0) + IO.succeed(1);
+// Result: IO<[number, number]>
+```
+
+**Enhanced Do Syntax**: Made effect-based programming more intuitive by introducing a clean, imperative-style syntax for monadic computations:
+
+```typescript
+// Traditional Effect code
```
IMax153, line 128:
```suggestion
// Traditional Effect code (prior to the removal of the `_` adapter)
```

content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * @tsplus global
+ */
+import { Option } from "@tsplus/stdlib/data/Option/definition";
+
+// Usage: no imports needed, types available globally
+const processData = (items: Chunk<string>) => {
+  return items.map(item => Option.some(item));
+};
```
IMax153, line 171: Indicate this is in a separate file somehow.

content/src/content/docs/blog/ts-plus-postmortem.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+**Complete Ecosystem Rewrite**: Perhaps most significantly, we rewrote the entire Effect ecosystem using TS+. This meant porting thousands of lines of carefully crafted functional programming abstractions, rebuilding our core libraries, and ensuring that all the complex type-level computations still worked correctly with our enhanced compiler.
+
+**Production Adoption**: The experiment reached a point where we had production users depending on alpha releases. Real projects, real applications, real teams betting their development workflow on our experimental fork. This level of adoption validated that TS+ wasn't just a theoretical exercise—it was solving genuine developer pain points.
+
+The scale of our commitment demonstrated our conviction that language tooling could be dramatically improved. We weren't just building a proof-of-concept; we were building a complete alternative development experience.
```
IMax153, line 207:
```suggestion
The scale of our commitment demonstrated our conviction that language tooling could be dramatically improved. We weren't just building a proof-of-concept; we were building a completely different development experience.
```

IMax153 (commented):
A global comment - the Bold Header + paragraph of content makes this feel AI generated. Even if it is only partially AI generated or AI collaborated on, I would suggest removing the bold heading + paragraph structure and instead use long form paragraphs like you've done in previous blogs.

PR #1411, opened 2026-07-29
fix: play `?code` silently ignored

f15u:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Opening a playground link like `/play?code=<snippet>` sometimes loaded whatever was last saved in `localStorage` instead of the linked snippet. Refreshing didn't help. Clearing `localStorage` didn't help either — it got rewritten within 2 seconds by the autosave loop.

## Root cause

`importAtom` resolved the initial workspace as `#hash → ?code → localStorage autosave → default`, all wrapped in a single blanket `Effect.catch`. If the hash branch ran and failed (bad/expired share link, network error), the catch swallowed it and fell straight to the autosave — `?code` was never even read. Any URL fragment (including `#_top` from the skip-link) was enough to trigger this.

Separately, the autosave loop ran unconditionally every 2s and would immediately persist whatever was just imported from a link, indistinguishable from a real edit — which is why clearing `localStorage` while the tab was open appeared to do nothing.

## Fix

- `atoms/import.ts`: each source (hash, `?code`, autosave, default) is now tried independently. A failing source is logged and falls through to the next one instead of aborting the whole chain.
- Autosave now skips writing while the workspace still matches what was just imported, so opening a link no longer overwrites a saved session before you've typed anything.
- Added a fallback error UI (`.onFailure`) with a "Reset playground" action for the now-rare case of a genuine defect, since the loader previously rendered nothing on failure.
- Fixed the "Open in Playground" button emitting `/play/?code=` (trailing slash) — 404s under this project's `trailingSlash: "never"` config; only masked in prod by a redirect.


src/features/playground/atoms/import.ts
```diff
@@ -25,6 +25,26 @@ const codeAtom = Atom.searchParam("code", {
   schema: Schema.StringFromBase64Url.pipe(Schema.check(Schema.isNonEmpty())),
 })
 
+/**
+ * Whether `snapshot` still looks like an untouched `baseline` (same file
+ * count, and `src/main.ts` unedited). Used to avoid persisting an import
+ * (default workspace, or whatever the workspace was imported as) to the
+ * autosave before the user has actually changed anything.
+ */
+function isUnchangedFrom(snapshot: Workspace, baseline: Workspace): boolean {
```
IMax153: I'm pretty sure `Workspace` implements `Hash` / `Equal` - can we not use those to perform this equivalence check?

Or is this doing something different / custom?

f15u: `Workspace` uses just `Hash`; this was needed. 

I agree that `Equal` is more idiomatic and we should use it.

IMax153 (approved):

PR #1416, opened 2026-07-30
refactor: make blog feature and content improvements

f15u:
## Type

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

These are bugs fixed:
- Featured post no longer duplicated in both hero and grid
- Prev/Next orientation was inverted (newest-first list, but Previous pointed to the newer post)
- Trailing slashes on RSS links / share links, conflicting with `trailingSlash: "never"`
- Author names weren't XML-escaped in the RSS feed
- Dates formatted in build-host timezone instead of UTC (off-by-one-day risk)
- Client hydration mismatch on `/blog?category=x` (category read from window during the render instead of after)
- Two client islands serializing overlapping post data → collapsed into one (`BlogIndex`)
- `getCollection` was called once per static page (171×) instead of once
- Dead `readingTime` schema field removed; `"release" → "releases"` typo in suggested tags fixed

## Related

- Related Issue #
- Closes #


f15u:
@IMax153 I've split up the MR, so this is focused just on the refactor without any other commit.

src/features/blog/content.test.ts
IMax153, line 1: This test feels useless.

IMax153 (commented):
Can the tests live in a separate top-level test dir? I don't like co-locating them in projects.

src/features/blog/repository.ts
```diff
@@ -0,0 +1,54 @@
...
+      tags: Effect.succeed(fixtures.tags ?? []),
+      authors: Effect.succeed(fixtures.authors ?? []),
+    })
+
+  static readonly layerFailing = (error: BlogContentError) =>
+    Layer.succeed(this, {
+      posts: Effect.fail(error),
+      tags: Effect.fail(error),
+      authors: Effect.fail(error),
+    })
```
IMax153: These layers should probably just be defined in the tests / where they are needed.

PR #1479, opened 2026-08-13
Add Effect v4/v3 version toggle to the playground

tim-smart:
Adds an Effect version toggle to the playground so users can switch between v4 (rc) and v3, following [Mirela's prototype](https://mirelaprifti.github.io/landing/play/).

Closes EFF-638

## What changed

- **Toolbar above the editor.** Reset and Share no longer float over editor code; they moved into a toolbar row above the editor together with the new version control (segmented `v4 (rc)` / `v3`, labeled "Effect version for this playground"). The row wraps on narrow viewports, so nothing covers code on desktop or mobile.
- **Per-version default workspaces.** v4 installs `effect@rc` + `@effect/platform-node@rc` and has its own v4-compatible `main.ts` and DevTools starter files; v3 keeps the existing template. Switching (with a confirmation dialog, since it discards current code) swaps the workspace handle, remounts, and re-runs the install. The active version is derived from the workspace's `package.json`, including the published `rc` and `beta` v4 tags plus exact v4 versions and ranges.
- **v4 DevTools bridge.** The v4 starter uses `NodeSocket.layerNet({ port: 34437 })` from `@effect/platform-node`. The playground-side decoder accepts both the native v4 and legacy v3 wire encodings (native first), so the Trace Viewer works on both versions.
- **Links.** The selected version persists in localStorage. `?code` links accept an optional `version` parameter, and docs "Open in playground" buttons pass the page's docs version. Unversioned `?code` links deliberately remain v3 because links created before this toggle contain v3 source.
- **Autosave compatibility.** Invalid user-edited `package.json` content no longer kills the autosave loop, and the version-switch staleness guard uses each workspace handle's immutable initial version so later package edits do not silently disable autosave.

## Verified

- Default playground boots v4, installs `effect 4.0.0-rc.108`, and runs the v4-specific starter
- Switching to v3 installs `effect 3.22.1`, and both versions send spans to the Trace Viewer
- Toggle never overlaps editor code at desktop or mobile breakpoints
- Version choice persists across reloads without resurrecting the outgoing workspace
- Call-site tests cover legacy unversioned `?code` links as v3, explicit v4 links, and both matching and mismatched autosave handle identities
- Mutation checks confirmed those tests fail if either link policy is removed, autosave uses the live package-derived version, or the mismatched-handle stale guard is removed
- `vp run check` passes with no formatting, lint, type, or Astro diagnostics
- `vp test` passes all 27 tests across 13 files


fubhy (dismissed):
@tim-smart This needs a different code starting point for v4

tim-smart:
> @tim-smart This needs a different code starting point for v4

Not sure what you mean. This changes the package.json etc.?

apps/web/src/features/playground/domain/workspace.ts
```diff
@@ -443,21 +475,73 @@ export const DevToolsLayer = DevTools.layerSocket.pipe(
...
+// web-stream view instead.
+const devToolsV4 = makeFile(
+  "DevTools.ts",
+  `import { Effect, Layer } from "effect"
+import { DevTools } from "effect/unstable/devtools"
+import { Socket } from "effect/unstable/socket"
+import * as Net from "node:net"
+import { Duplex } from "node:stream"
+
+const NetSocket = Layer.effect(
```
tim-smart: This should use NodeSocket from @effect/platform-node

fubhy:
> > @tim-smart This needs a different code starting point for v4
> 
> Not sure what you mean. This changes the package.json etc.?

Nevermind. I read past that somehow.

PR #1482, opened 2026-08-13
Fix docs search version preferences

tim-smart:
## Summary

- initialize docs search from the current `/docs/v3` or `/docs/v4` route whenever the dialog opens
- persist the selected version with each recent search and restore it when that search is selected
- preserve legacy query-only history as v3 entries and display version badges in the recent list
- add focused regression coverage for route detection, version persistence, deduplication, and legacy history

## Root cause

The dialog version atom was always initialized to v3, and recent searches stored only a query string. As a result, neither the current docs route nor a historical search could provide the version used to filter results.

## User impact

Search now opens against the docs version being viewed, and selecting a recent search restores the matching version switch and result set.

## Validation

- `nix develop -c pnpm exec vp run check`
- `nix develop -c pnpm exec vp test` (31 tests)
- `nix develop -c pnpm exec vp run build`
- local Playwright verification across `/docs/v4`, `/docs/v3`, and v4 recent-search restoration

Closes EFF-652

### DIFF
--- a/apps/web/src/components/navigation/SearchDialog.tsx
+++ b/apps/web/src/components/navigation/SearchDialog.tsx
@@ -52,11 +52,17 @@ import {
   type BlogSearchResult,
   type DocumentationSearchResult,
 } from "@/services/search/domain"
+import {
+  addRecentSearch,
+  type RecentSearch,
+  RecentSearches as RecentSearchHistory,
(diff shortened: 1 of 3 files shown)


f15u:
Superseded by #1483

PR #1522, opened 2026-09-03
Prevent theme toggle layout shift

Andarist:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

I noticed an annoying layout shift caused by the theme toggle switch not being available until client-side hydration. This PR makes both icons rendered on the server and allows Astro to quickly decide which one to render at initialization time (it toggled `.dark` class on the `html` element). The icon's visibility state gets toggled with the CSS classes now and doesn't need an atom subscription.

## Related

N/A


f15u:
Thank you for the PR!

PR #1542, opened 2026-09-09
fix: animate FAQ accordion on open and close

### LINKED ISSUE #1541: Landing page FAQ accordion snaps open and closed instead of animating
The FAQ accordion on the landing page is styled to expand and collapse over 200ms, but the panel jumps straight to its new height instead.

## Steps to reproduce

1. Open https://effect.website/ and scroll down to "Frequently asked questions"
2. Click a question, observe the panel expands (good)
3. Click it again, the panel vanishes instantly instead of collapsing (bad)
4. Click a third time, it reappears instantly (bad)

mrlubos:
## Type

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The FAQ accordion on the landing page snaps instead of animating.

The panel animation is a `grid-template-rows: 0fr → 1fr` transition on a wrapper `<div>` inside the `<details>`. In browsers that support `::details-content`, a closed `<details>` gets `content-visibility: hidden` on that pseudo-element, which skips rendering of everything beneath it. The transition therefore never advances while the panel is closing: the row stays pinned at its open height, which in turn leaves the next open with nothing to interpolate from.

This moves the transition onto `::details-content` itself, whose own box is always styled, and adds `allow-discrete` so the flip to `content-visibility: hidden` lands at the end of the collapse rather than on its first frame.

The new rules sit behind `@supports selector(::details-content)` and the existing wrapper is left in place, so browsers without support fall back to exactly the markup and CSS that ship today. The only change outside the `@supports` block is two added class names, which nothing else on the page styles.

## Related

- Closes #1541

### DIFF
--- a/apps/web/src/components/landing/blocks/FrequentlyAskedQuestions.astro
+++ b/apps/web/src/components/landing/blocks/FrequentlyAskedQuestions.astro
@@ -35,7 +35,7 @@ const FAQS: FaqItem[] = [
 <div class="space-y-4">
   {
     FAQS.map(({ question, answer }) => (
-      <details class="group rounded-md border border-zinc-700 hover:border-zinc-600 hover:bg-zinc-900/50">
+      <details class="faq-item group rounded-md border border-zinc-700 hover:border-zinc-600 hover:bg-zinc-900/50">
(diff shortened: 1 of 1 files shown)


f15u:
Thank you for the PR!

